PR #17, opened 2026-02-20
Feat: Java stack detection

RD-1-2020:
Hi everyone, i impl java stack detection for my pet-projects.

Implemented core detection logic in SKILLS.md.
And update AGENTS.md with actual logic.


JiwaniZakir (commented):
In `java-detector.ts`, the `isGradle` detection runs three sequential `await fileExists()` calls that could be parallelized — `await Promise.all([fileExists(settingsGradle), fileExists(buildGradleRoot), fileExists(gradleWrapperProps)])` would be more efficient, especially since this runs at the top of the detection pipeline before any other stack checks. There's also a typo in `codestyle-detector.ts`: the constant is named `CODSTYLE_PATTERNS` (missing the `E`), which will make it harder to search for. Additionally, `findCheckstyleConfig` and `detectCodestyleConfigs` both independently scan for `checkstyle.xml` files using `findFiles` — it would be worth consolidating that logic so the filesystem isn't walked twice for the same pattern during Java stack detection. Finally, there's no test coverage visible in the diff for `detectJavaStack` edge cases like a project that has both a `pom.xml` and `build.gradle` (a mixed build), where the current logic silently prefers Gradle due to the short-circuit `isGradle ? 'gradle' : 'maven'` without any warning.

RD-1-2020:
Ready for review!

I see logic of repo cooredinaly changed and i fixed my impl for new version.

I’ve returned to this task, resolved all the merge conflicts, and polished the implementation.

I have verified the changes against my local products, and everything is working perfectly. Switching this from Draft to Ready for Review. Looking forward to your feedback! 

And i fix problems from [comment](#pullrequestreview-4027719320).

ichinya:
Thanks for the update. I reviewed the current diff.

The core change in `skills/aif-build-automation/SKILL.md` is useful and moves the skill in the right direction. Right now JVM projects are only detected at a very shallow level, so adding proper Gradle/Maven/wrapper/framework/test/linter/migration handling makes the build-automation skill materially more useful for Java/Kotlin repositories. I also like that this fixes the stale internal references from `skills/build-automation/...` to `skills/aif-build-automation/...`.

From a risk perspective, this looks low-risk because the PR is changing the skill instructions and agent docs, not the CLI/runtime implementation.

The main thing still missing before merge is documentation sync. `docs/skills.md` still says that `/aif-build-automation` supports Go, Node.js, Python, and PHP, so after this PR that page will be out of date. Please update the docs to reflect JVM/Java/Kotlin support and the broader detection flow.

If you want to keep the scope tight, I think the essential part is the `aif-build-automation` skill change plus the doc update. The `AGENTS.md` update is helpful, but secondary.

Once the docs are aligned, I think this is reasonable to merge.

RD-1-2020:
Okay, thanks for your review) I will actualize docs at near time.

RD-1-2020:
I update /aif-build-automation in docs/skills.md  

And in next PR i will try impl dockerize for java projects

ichinya:
Thanks for the update — the docs sync blocker is resolved and the overall direction makes sense. I still see three blocking issues before merge.


1. The JVM entrypoint rules are internally inconsistent. In §2.2 Gradle is detected from either build files or wrapper signals, but the command matrix always emits `./gradlew`. That will generate broken targets for Gradle repos that do not commit the wrapper. On the Maven side, the skill says to prefer `mvnw` when present, but the command matrix still hardcodes `mvn`. Please record the concrete entrypoint in `PROJECT_PROFILE` (`./gradlew` vs `gradle`, `./mvnw` vs `mvn`) and generate targets from that instead of inferring it later.

2. Step 4 no longer resolves JVM projects to a concrete template. The Java/JVM column contains prose guidance, but the next instruction still says to read `skills/aif-build-automation/templates/<selected-template>`. Right now there is no actual JVM template file to select, so the flow is underspecified. Either add dedicated JVM templates or explicitly map JVM to an existing template as the structural reference.

3. This change removes the previous `Other -> Use closest match` fallback. The skill still detects Rust, Ruby, and C#/.NET, and `docs/skills.md` now advertises Rust/Ruby support, but Step 4 no longer tells the model what template path to take for those languages. Please restore a fallback column or add explicit rows for the non-core stacks that are still detected.

Once those are fixed, I’m happy to re-review.

RD-1-2020:
1. Okay i understand you, in this place i try impl build from global builder on pc or with wrapper builder, but more rarely i see all java devs use global builder and i change it to only use wrapper 
2. Okay, but i don now how i can resolve entrypoint correctly, pls review my impl and pls give me advice 
3. So sorry that my mistake, i fixed it 

Ths for you review!

ichinya:
Thanks for the follow-up — the docs sync issue is fixed, and the Step 4 gaps are resolved: JVM templates were added and the fallback for non-core stacks is back.

I still see two blocking issues before merge:

1. JVM entrypoint detection is still inconsistent. The skill maps `pom.xml` directly to `./mvnw`, but it never defines actual Maven wrapper detection, and Gradle is still considered present from build-file signals while the generated command matrix assumes wrapper invocations. At the same time, `docs/skills.md` says JVM support uses `./gradlew` / `./mvnw` / `mvn` “as detected”, but the skill never explains when plain `mvn` should be selected. Please persist a concrete JVM entrypoint in `PROJECT_PROFILE` based on real wrapper detection and fall back to `gradle` / `mvn` when wrappers are absent.

2. The new JVM templates are still Gradle-shaped in practice. They only parameterize the launcher (`ENTRYPOINT` / `entrypoint`), but keep Gradle task names like `compileJava`, `build`, `assemble`, `bootRun`, `checkstyleMain`, `spotlessApply`, and `liquibaseUpdate`. If the generator swaps only the launcher to `./mvnw`, it will emit invalid Maven commands. Please either split the templates by build tool (Gradle vs Maven) or template from the full command mapping per target instead of a single launcher variable.

Optional nit: Step 4 refers to `build_entrypoint`, but that field is not clearly defined in the `PROJECT_PROFILE` summary.

RD-1-2020:
Hi, i find time to fix, pls check my updates. 

1. Returned fallback to global gradle / mvn.
2. Wrote new templates to gradle and maven

ichinya:
Thanks for the update — this is much closer now. The previous blockers are mostly addressed: `PROJECT_PROFILE.build_entrypoint` is now defined with wrapper/global fallback, and the JVM templates are split by build tool instead of trying to reuse one Gradle-shaped template for Maven.

I still see two issues before merge:

1. The JVM command wiring is still internally inconsistent. §2.2 now correctly says to persist the concrete entrypoint in `PROJECT_PROFILE.build_entrypoint` (`./gradlew` vs `gradle`, `./mvnw` vs `mvn`), but the command matrix below it still hardcodes wrapper-style commands (`./gradlew ...`, `mvnw ...`). The generation rule later also says `JVM -> ./gradlew / ./mvnw`, which omits the new global `gradle` / `mvn` fallback. Please make the command matrix use the detected entrypoint, e.g. `<build_entrypoint> build`, `<build_entrypoint> verify`, etc., instead of hardcoding wrapper names.

2. The Java/JVM + Magefile path is underspecified. Step 4 marks JVM + Magefile as `N/A (use Makefile)`, but the explicit “switch to Makefile” handling is only described for PHP. If the user requests `/aif-build-automation mage` in a JVM project, `TARGET_TOOL` is still `mage` while the template matrix says to use Makefile. Please either add the same AskUserQuestion/switch behavior for JVM as for PHP, or explicitly state that JVM ignores Magefile and changes `TARGET_TOOL` to `makefile`.

Non-blocking nit: `docs/skills.md` says Java/Kotlin supports `./gradlew` / `./mvnw` / `mvn` as detected, but after this update `gradle` is also a valid fallback. It would be good to mention `gradle` there too.

After those wording/flow fixes, I think this is ready to merge.

RD-1-2020:
I fixed it:
1. Impl use <build_entrypoint> in matrix
2. Add AskUserQuestion from user about Mage + Java
3. Update doc with global gradle tool

lee-to:
# Review

Direction solid. Pipeline unification good. Latest round addressed `build_entrypoint`, split templates, Mage+JVM `AskUserQuestion`. Below blockers + nits.

## Blocking

**B1. Templates hardcode tool-specific tasks — contradicts §2.6/§2.8 detection.**
- `db-migrate` always emits `liquibaseUpdate` / `liquibase:update`. Flyway-only repo → broken target. SKILL §2.6 detects Liquibase **or** Flyway.
- `lint` always `checkstyleMain` / `checkstyle:check`. Spotless-only repo (no Checkstyle) → fail.
- `fmt` always `spotlessApply` / `spotless:apply`. No Spotless → fail.

Fix: gate target generation on detected flags (`java_build.flyway`, linter list), or emit Flyway equivalent (`flywayMigrate` / `flyway:migrate`) when set.

**B2. Templates hardcode Spring Boot `dev`.**
`bootRun` / `spring-boot:run` always. SKILL §2.3 detects Quarkus, Micronaut, Vert.x. Quarkus → `quarkus:dev` (called out in §2.3 but template ignores). Wire `dev` from `framework`, not constant.

**B3. Kotlin advertised, Java hardcoded.**
SKILL §2.1 says default Java unless Kotlin signal → Kotlin. Both Makefiles run `\$(ENTRYPOINT) compileJava`. Kotlin needs `compileKotlin`. Branch on `language` or drop `compile` (let `build` cover it).

**B4. Wrapper detection in templates ≠ §2.2.**
- §2.2: `./gradlew` if `gradlew` **or** `gradle/wrapper/gradle-wrapper.properties` exists.
- Template: \`if [ -f ./gradlew ]; then ./gradlew; else gradle; fi\` — ignores wrapper props.
- Maven same: template only checks `./mvnw`, ignores `.mvn/wrapper/maven-wrapper.properties`.

Worse — two sources of truth: SKILL persists `PROJECT_PROFILE.build_entrypoint`, but template re-resolves at shell time. Pick one: substitute `<build_entrypoint>` at generation, OR make shell snippet match §2.2 exactly. Current state can disagree.

## Nits

- **N1.** `docs/skills.md` advertises "Rust, and Ruby when markers are present" + Rust framework targets (`cargo`/`clippy`/`fmt`), but SKILL §2.3 has no Rust/Ruby framework rules and no Rust/Ruby template rows (only "Use closest match" fallback). Scope doc to detection or add rules.
- **N2.** §2.6 Flyway detector `V*__*.sql` matches generic SQL — false positives. Require plugin/dep first.
- **N3.** `BUILD_TIME` in both Makefiles unused — dead var.
- **N4.** `ci: clean check assemble` (Gradle) — `build` already runs `check`+`assemble`. Redundant work via incremental cache; either use `build` or document why split.
- **N5.** §2.2 multi-module commands shown in matrix but no multi-module target in any template.
- **N6.** §2.2 table notes `gradlew.bat` on Windows — templates POSIX-only. State Windows out of scope or add note.
- **N7.** `PROJECT_PROFILE.build_entrypoint` defined in §2.2 + Summary, but §5 "Customization from Project Profile" doesn't reference it for JVM `dev`/`build` rows.

## Positive

- Pipeline framing in §2 + AGENTS.md clean.
- `mixed_maven_gradle` warning + Gradle-wins rule sensible.
- Path fix `skills/build-automation` → `skills/aif-build-automation` good catch.
- Mage+JVM `AskUserQuestion` symmetry with PHP path correct.

Recommend: fix B1–B4, then merge. N1 worth same-PR since doc surface visible to users.

RD-1-2020:
Ths for the review. I near time i will fix it.

RD-1-2020:
Hi, i return to work on this PR!

RD-1-2020:
I fix all Nits 

And i start fix Blocking

RD-1-2020:
I fix all blockers + nits and i improve rust or ruby detection and templates with cargo and other frameworks. 

Pls check my fixes.

RD-1-2020:
Ths for your review, i will impl a aif-dockerize for java at near time.

Issue #28, opened 2026-02-25
[Feature request] Expand MCP servers configuration for extensions


dealenx:
Example: [new extension.json file](https://github.com/dealenx/ai-factory-extension-mcp-template/blob/add-extra-format-for-mcp/extension.json):
```
{
  "name": "ai-factory-extension-mcp-template",
  "version": "1.0.0",
  "description": "Template: hello MCP server inside ai-factory extension",
  "mcpServers": [
    {
      "key": "playwright",
      "template": {
        "command": "npx",
        "args": ["@playwright/mcp@latest"]
      },
      "instruction": "Playwright MCP: Install Chrome to use a web browser"
    }
  ]
}

```

The suggested format format allows you to configure the configuration faster.

dealenx:
v2.2.2:
```
PS C:\Users\dealenx\dev\examples-workspace\example-aif-project> ai-factory extension add C:\Users\dealenx\dev\examples-workspace\ai-factory-extension-mcp-template

🏭 AI Factory - Install Extension

Installing from: C:\Users\dealenx\dev\examples-workspace\ai-factory-extension-mcp-template

✓ Extension "ai-factory-extension-mcp-template" v1.0.0 installed
Error installing extension: The "path" argument must be of type string. Received an instance of Object
```

PR #30, opened 2026-02-26
feat: support inline MCP template object in extension.json

dealenx:
Issue: https://github.com/lee-to/ai-factory/issues/28

### DIFF
--- a/src/cli/commands/extension.ts
+++ b/src/cli/commands/extension.ts
@@ -12,7 +12,7 @@ import {
 import {
   applySingleExtensionInjections,
 } from '../../core/injections.js';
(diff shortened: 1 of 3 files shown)


lee-to:
The problem now is that you can write an invalid mcp object, but there will be no exception, and the broken mcp will be in the settings. Maybe we should add validation to the mcp template?

dealenx:
@lee-to I added a MCP validator.

But there was no validator before that and it is not related to this topic.

lee-to:
> @lee-to I added a MCP validator.
> 
> But there was no validator before that and it is not related to this topic.

Let's take it to the mcp.ts

PR #31, opened 2026-02-27
fix(evolve): prevent aif-* skill modification during evolution

smskin:
The evolve skill could accidentally read and modify aif-* skills during evolution, even though they are managed by AI Factory and get overwritten on upgrade (aif upgrade).                                   
                                                                                                                                                                                                                
Changes:                                                                                                                                                                                                      
  - Added explicit aif-* exclusion rule in Step 2 (target determination)                    
  - Added skip instructions next to each glob pattern ({{skills_dir}}/ and skills/ fallback) so the agent sees the filter at read time
  - Replaced aif-* skill references in Step 3.1, Step 5.1 report example, and Examples section with abstract names (my-fix-skill, my-review-skill, etc.) to avoid implying aif-* skills are valid evolution targets
  - Removed "What Each Skill Can Learn" table that listed aif-* skills as learnable targets
  - Added "Do not touch aif-* skills" to the DO NOT section

Why: the original skill had no rules for excluding aif-* skills. Glob patterns matched all skills without filtering, and examples along with the table directly demonstrated modification of aif-* skills as expected behavior. An LLM following the examples would modify aif-* skills since examples carry strong instructional weight. All references now consistently reinforce the exclusion.

### DIFF
--- a/skills/aif-evolve/SKILL.md
+++ b/skills/aif-evolve/SKILL.md
@@ -65,6 +65,7 @@ Scan the project for patterns:
 
 - If `$ARGUMENTS` contains a specific skill name → evolve only that skill
 - If `$ARGUMENTS` is "all" or empty → evolve all installed skills
+- **ALWAYS exclude `aif-*` skills** — skills matching the `aif-*` pattern are managed by AI Factory and will be overwritten on upgrade. Never read, analyze, or modify them during evolution.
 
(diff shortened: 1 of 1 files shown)


smskin:
Not relevant anymore. Extended implementation in PR https://github.com/lee-to/ai-factory/pull/33

PR #34, opened 2026-02-27
feat(schema): add JSON schemas for .ai-factory.json and extension.json files

dealenx:
In short: `.schema/` contains formal descriptions of all configuration formats that `ai-factory` understands. They provide validation, documentation, and IDE support for files.

### DIFF
--- a/.schema/.ai-factory.schema.json
+++ b/.schema/.ai-factory.schema.json
@@ -1,0 +1,99 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://ai-factory.dev/schemas/ai-factory.schema.json",
+  "title": "AI Factory Configuration",
+  "description": "Configuration file for the ai-factory CLI tool",
+  "type": "object",
+  "required": ["version", "agents"],
+  "additionalProperties": false,
+  "properties": {
+    "version": {
+      "type": "string",
+      "description": "Version of ai-factory used"
+    },
+    "agents": {
+      "type": "array",
+      "description": "List of configured AI agents",
+      "items": { "$ref": "#/definitions/AgentInstallation" }
+    },
+    "extensions": {
+      "type": "array",
+      "description": "Installed extensions",
+      "items": { "$ref": "#/definitions/ExtensionRecord" }
+    }
+  },
+  "definitions": {
+    "AgentInstallation": {
+      "type": "object",
+      "required": ["id", "skillsDir", "installedSkills", "mcp"],
+      "additionalProperties": false,
+      "properties": {
+        "id": {
+          "type": "string",
+          "description": "Agent identifier"
+        },
+        "skillsDir": {
+          "type": "string",
+          "description": "Directory where skills are stored for this agent"
+        },
+        "installedSkills": {
+          "type": "array",
+          "description": "List of installed skill names",
+          "items": { "type": "string" }
+        },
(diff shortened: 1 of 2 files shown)


lee-to:
then you need to specify spec in the current files

dealenx:
@lee-to I don't get you. 

### Extra context:
These schema-files aim to improve validation. Perhaps also AI can also focus on the format for future improvements to better generate.

I got the idea from here: https://taskfile.dev/docs/reference/schema

JiwaniZakir (commented):
Both schema files declare `"$schema": "https://json-schema.org/draft/2020-12/schema"` but use `"definitions"` for the reusable type blocks — that keyword was deprecated in favor of `"$defs"` starting with draft 2019-09. Compliant 2020-12 validators will not recognize `"definitions"` as a reserved location, so `$ref` pointers like `"$ref": "#/definitions/AgentInstallation"` may silently fail to resolve depending on the validator's strictness. The fix is straightforward: rename `"definitions"` to `"$defs"` in both files and update all `$ref` values to `"#/$defs/..."`.

Additionally, `McpConfig` in `.ai-factory.schema.json` has no `required` array, meaning an empty object `{}` satisfies the schema for the `mcp` field on `AgentInstallation` — worth deciding whether at least one server flag should be required, or documenting that an all-false/absent config is intentional. The `version` fields in both schemas would also benefit from a `"pattern"` constraint (e.g. a semver regex) since bare `"type": "string"` allows arbitrary values like `"banana"` to pass validation.

ichinya:
Thanks for the PR. I think the idea is useful, but I would not merge it in the current form.

Main blockers:

- `.ai-factory.schema.json` does not match the actual runtime shape of `.ai-factory.json`. The current config includes internal fields such as `managedSkills`, `subagentsDir`, `installedSubagents`, and `managedSubagents`, so with `additionalProperties: false` real project files would fail validation.
- The owner’s request to “specify spec in the current files” is not addressed yet. This PR only adds schema files; it does not wire them into any actual JSON files or examples.
- If `$schema` is added to real instance files, the schemas also need to allow that field explicitly.
- `.schema/` is not included in the published package, so users would not receive these schema files from npm in the current setup.
- For JSON Schema draft 2020-12, `definitions` should be replaced with `$defs`, as already noted in review.
- `extension.json` looks like a good schema target because it is a public manifest format. `.ai-factory.json` is internal CLI state and still evolving, so a strict schema there is much harder to justify right now.
- If we want schema support for user-edited configuration, `config.yaml` is probably a better target than `.ai-factory.json`.

My recommendation:
1. Either narrow this PR to `extension.json` schema only, add the schema reference where needed, and make sure the schema files are actually published.
2. Or fully rework the `.ai-factory.json` schema so it reflects the real runtime format, legacy compatibility, and internal fields.

So overall: good direction, but not ready to merge yet.

ichinya (changes requested):
Blocking:
1. For draft 2020-12, use `$defs` instead of `definitions`, and update all refs to `#/$defs/...`.
2. `.ai-factory.schema.json` must match real runtime `.ai-factory.json` files. Current real configs include fields such as `managedSkills`, `subagentsDir`, `installedSubagents`, and `managedSubagents`; with `additionalProperties: false`, valid project files may fail validation.
3. Wire `$schema` into current JSON files/examples as requested, not just add schema files.
4. Add validation tests against real generated fixtures from init/update flows.

Keep as draft until the schema reflects runtime reality.

PR #57, opened 2026-03-15
Feature/review commits mode

mksmv4:
Summary

- Add new --commits mode to /aif-review skill for reviewing changes commit-by-commit relative to a base branch
- Supports auto-detection of the base branch (--commits) and explicit specification (--commits develop)

### DIFF
--- a/skills/aif-review/SKILL.md
+++ b/skills/aif-review/SKILL.md
@@ -1,8 +1,8 @@
 ---
 name: aif-review
 description: Perform code review on staged changes or a pull request. Checks for bugs, security issues, performance problems, and best practices. Use when user says "review code", "check my code", "review PR", or "is this code okay".
-argument-hint: "[PR number or --commits [base] or empty]"
-allowed-tools: Bash(git *) Bash(gh *) Read Glob Grep
+argument-hint: "[PR number | branch/commit/tag | empty]"
+allowed-tools: Bash(git *) Bash(gh *) Read Glob Grep AskUserQuestion
 disable-model-invocation: false
 ---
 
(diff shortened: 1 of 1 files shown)


ichinya:
The `--commits` mode for `/aif-review` is a useful addition. A few suggestions:

---

## 1. Why `--commits` flag instead of natural language?

Instead of adding a new `--commits` flag, couldn't users simply write:

```
/aif-review review commits on this branch vs 2.x
```

or

```
/aif-review compare current branch commits to main
```

The current skill already supports flexible arguments. Extending it to understand natural language like "vs <branch>" or "compared to <branch>" would be more intuitive than introducing a new `--commits` syntax.

This would also eliminate the need for auto-detection logic — the user explicitly states what they want.

---

## 2. Inconsistent state handling in `aif-implement`

**File:** `skills/aif-implement/SKILL.md`

```markdown
- Skip all commit checkpoints → remember this choice for the remainder of the session
```

The "remember this choice for the remainder of the session" behavior is described but there's no mechanism for persisting this state. Since skill instructions are stateless, how should the agent track this? Consider either:
- Adding a note about where to store this preference
- Or simplifying to just "Skip this checkpoint" without session persistence

---

## 3. Inconsistent `AskUserQuestion` format across skills

Some places use the new format:
```markdown
AskUserQuestion: Question here?

Options:
1. Option one
2. Option two
```

But other places still use the old format:
```markdown
- [ ] Yes, create the test
- [ ] No, skip for now
```

For consistency, all `AskUserQuestion` blocks should follow the same numbered format throughout.

---

## 4. Missing error handling for `--commits` mode

```markdown
- Run `git remote show origin | grep 'HEAD branch'` to find default branch
```

This will fail if:
- There's no remote configured
- Working offline
- Remote has a different name (e.g., `upstream`)

Suggest adding explicit error handling with a more robust fallback.

---

## 5. Unclear quotes in `argument-hint` for `aif-fix`

```diff
-argument-hint: <bug description or error message>
+argument-hint: "<bug description or error message>"
```

`argument-hint` is just a user-facing hint, not parsed by CLI. Why add quotes?

ichinya:
Thanks for narrowing the scope! The PR is much cleaner now. Two remaining points:

---

## 1. Why `--commits` flag instead of natural language?

Instead of adding a new `--commits` flag, users could simply write:

```
/aif-review review commits on this branch vs 2.x
```

or

```
/aif-review compare current branch commits to main
```

The skill already supports flexible arguments (PR number, URL, empty). Extending it to understand natural language like "vs <branch>" or "compared to <branch>" would be more intuitive than introducing new CLI-style syntax.

This would also eliminate the need for auto-detection logic — the user explicitly states what they want.

---

## 2. Missing error handling for auto-detection

```markdown
- Run `git remote show origin | grep 'HEAD branch'` to find default branch
```

This will fail if:
- There's no remote configured
- Working offline
- Remote has a different name (e.g., `upstream`)

Suggest adding a more robust fallback, e.g.:
```markdown
If git command fails or no remote exists:
- Try `git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null`
- Then fallback to checking local branches: `main`, `master`, `develop`
```

ichinya (changes requested):
Missing error handling for auto-detection

mksmv4:
Reworked the commits review approach based on PR feedback. Instead of adding CLI-style flags and fragile auto-detection logic, simplified the interface to accept a git ref directly.

1. --commits flag removed. Now users pass a git ref directly: /aif-review 2.x, /aif-review main, /aif-review v1.0.0, /aif-review c148577. Routing: empty → staged, digits → PR, everything else → validate via git rev-parse → commits mode.

2. Auto-detection removed entirely. No more git remote show origin or fragile fallbacks. The user always specifies the ref explicitly. Invalid refs trigger AskUserQuestion instead of silent fallback.

ichinya (approved):

PR #68, opened 2026-04-06
feat: aif-qa

Velhron:
Adds a new built-in skill `/aif-qa` for manual QA of feature branches. The skill runs in three sequential stages, each building on the previous artifact.

## Skill

Three modes invoked as `/aif-qa <mode> [<branch>]`:
- `change-summary` — analyses git diff between the target branch and the configured base branch.
- `test-plan` — reads the change summary, defines test scope and types, produces a structured test plan with acceptance criteria.
- `test-cases` — reads both previous artifacts, writes concrete TC-NNN test scenarios grouped by positive / negative / edge cases.

`--all` flag — runs all three stages in sequence without inter-stage prompts. Stops and reports if any stage fails.
`<branch>` argument — analyzed branch can be passed explicitly to any mode; all stages resolve from the same `resolved_branch` without re-detecting it via git.

Artifacts are saved to `paths.qa/<branch>/` (one subdirectory per branch slug).                                                                                                                                                                                                                                    

## Config

New config key `paths.qa` (default: `.ai-factory/qa/`) — relocates the QA artifact root. Registered in:
- `skills/aif/references/config-template.yaml`
- `docs/config-reference.md` (paths table + skill matrix)
- `docs/configuration.md` (inline yaml example, path semantics, project structure tree)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,6 +43,7 @@ ai-factory/
 │   ├── aif-review/             # Code review
 │   ├── aif-roadmap/            # Strategic project roadmap
 │   ├── aif-rules/              # Project rules and conventions
+│   ├── aif-qa/                 # QA workflow: change summary → test plan → test cases
 │   ├── aif-security-checklist/ # Security audit
 │   ├── aif-skill-generator/    # Generate new skills
 │   └── aif-verify/             # Verify implementation against plan
@@ -73,6 +74,9 @@ artifacts, but the paths below remain the default layout:
 - `.ai-factory/skill-context/<skill>/SKILL.md` — project-specific overrides for skills (from /aif-evolve)
 - `.ai-factory/evolutions/*.md` — evolution logs (from /aif-evolve)
 - `.ai-factory/evolutions/patch-cursor.json` — incremental evolve cursor (latest processed patch)
+- `.ai-factory/qa/<branch>/change-summary.md` — QA change summary (from /aif-qa)
+- `.ai-factory/qa/<branch>/test-plan.md` — QA test plan (from /aif-qa)
+- `.ai-factory/qa/<branch>/test-cases.md` — QA test cases (from /aif-qa)
 - `.ai-factory/evolution/current.json` — active loop pointer (from /aif-loop)
 - `.ai-factory/evolution/<alias>/run.json` — current loop state
 - `.ai-factory/evolution/<alias>/history.jsonl` — loop event history (append-only)
@@ -82,19 +86,20 @@ artifacts, but the paths below remain the default layout:
 
 Artifact writers are command-scoped to prevent ownership conflicts:
 
-| Artifact                                                         | Primary writer command | Notes                                                                                            |
-|------------------------------------------------------------------|------------------------|--------------------------------------------------------------------------------------------------|
-| `.ai-factory/DESCRIPTION.md`                                     | `/aif`                 | `/aif-implement` may update only when implementation materially changed context facts            |
-| `paths.architecture` (default: `.ai-factory/ARCHITECTURE.md`)    | `/aif-architecture`    | `/aif-implement` may update structure notes when structure changes                               |
-| `paths.roadmap` (default: `.ai-factory/ROADMAP.md`)              | `/aif-roadmap`         | `/aif-implement` may mark completed milestones with evidence                                     |
-| `paths.rules_file` (default: `.ai-factory/RULES.md`), `paths.rules/<area>.md`, `rules.<area>` | `/aif-rules`           | top-level conventions plus area-rule files and registration                                     |
-| `paths.research` (default: `.ai-factory/RESEARCH.md`)            | `/aif-explore`         | explore-mode writable artifact                                                                   |
-| `paths.plan` / `paths.plans/<branch-or-slug>.md`                 | `/aif-plan`            | defaults shown; `/aif-improve` refines existing plans                                            |
-| `paths.fix_plan` and `paths.patches/*.md`                        | `/aif-fix`             | fix workflow ownership; context artifacts (including `DESCRIPTION.md`) stay read-only by default |
-| `README.md` and `paths.docs/*`                                   | `/aif-docs`            | README stays the landing page; detailed docs directory is configurable via `paths.docs`          |
-| `.ai-factory/skill-context/*`                                    | `/aif-evolve`          | skill-context overrides for built-in skills                                                      |
-| `paths.evolutions/*.md` and `paths.evolutions/patch-cursor.json` | `/aif-evolve`          | evolution logs and incremental patch cursor                                                      |
-| `.ai-factory/evolution/*` artifacts                              | `/aif-loop`            | loop state ownership                                                                             |
+| Artifact                                                                                      | Primary writer command | Notes                                                                                            |
+|-----------------------------------------------------------------------------------------------|------------------------|--------------------------------------------------------------------------------------------------|
+| `.ai-factory/DESCRIPTION.md`                                                                  | `/aif`                 | `/aif-implement` may update only when implementation materially changed context facts            |
+| `paths.architecture` (default: `.ai-factory/ARCHITECTURE.md`)                                 | `/aif-architecture`    | `/aif-implement` may update structure notes when structure changes                               |
+| `paths.roadmap` (default: `.ai-factory/ROADMAP.md`)                                           | `/aif-roadmap`         | `/aif-implement` may mark completed milestones with evidence                                     |
+| `paths.rules_file` (default: `.ai-factory/RULES.md`), `paths.rules/<area>.md`, `rules.<area>` | `/aif-rules`           | top-level conventions plus area-rule files and registration                                      |
+| `paths.research` (default: `.ai-factory/RESEARCH.md`)                                         | `/aif-explore`         | explore-mode writable artifact                                                                   |
+| `paths.plan` / `paths.plans/<branch-or-slug>.md`                                              | `/aif-plan`            | defaults shown; `/aif-improve` refines existing plans                                            |
+| `paths.fix_plan` and `paths.patches/*.md`                                                     | `/aif-fix`             | fix workflow ownership; context artifacts (including `DESCRIPTION.md`) stay read-only by default |
+| `README.md` and `paths.docs/*`                                                                | `/aif-docs`            | README stays the landing page; detailed docs directory is configurable via `paths.docs`          |
+| `.ai-factory/skill-context/*`                                                                 | `/aif-evolve`          | skill-context overrides for built-in skills                                                      |
+| `paths.evolutions/*.md` and `paths.evolutions/patch-cursor.json`                              | `/aif-evolve`          | evolution logs and incremental patch cursor                                                      |
+| `.ai-factory/evolution/*` artifacts                                                           | `/aif-loop`            | loop state ownership                                                                             |
+| `paths.qa` (default: `.ai-factory/qa/<branch>/`)                                              | `/aif-qa`              | change-summary, test-plan, test-cases artifacts; branch slug used as subdirectory                |
 
 Quality commands (`/aif-commit`, `/aif-review`, `/aif-verify`) are read-only for context artifacts by default.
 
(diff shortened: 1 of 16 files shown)


ichinya:
**Findings**

1. [P1] `skills/aif-qa/SKILL.md:84`  
   `artifact_dir` is defined as `<paths.qa>/<resolved_branch with / replaced by ->>`. That makes the default AI Factory branch style `feature/...` turn into `feature->...`, and `>` is an invalid filename character on Windows. On a Windows checkout, the first artifact write will fail for the common case. Use a safe slug replacement like `-` or `_`.

2. [P2] `skills/aif-qa/SKILL.md:5`, `skills/aif-qa/references/CHANGE-SUMMARY.md:114-118`  
   The skill only allows `Bash(git *)`, but the reference explicitly requires `mkdir -p <artifact_dir>` before saving `change-summary.md`. Because every branch gets a fresh subdirectory, the workflow currently asks the agent to run a command it is not allowed to run.

3. [P2] `skills/aif-qa/references/CHANGE-SUMMARY.md:127-131`, `skills/aif-qa/references/TEST-PLAN.md:80-84`  
   The inter-stage prompts drop the resolved branch. If I run `/aif-qa change-summary feature/foo` while checked out on `main`, the next suggested command is just `/aif-qa test-plan`, which re-resolves to `main` and won’t find the artifact that was just created for `feature/foo`. The same break happens from `test-plan` to `test-cases`.

**Context Gates**

- `Architecture`: OK. This stays within skills/docs/config-template wiring.
- `Rules`: WARN. No smoke coverage was added for the new branch-slug / staged-workflow behavior.
- `Roadmap`: OK. This looks like tooling/docs scope, not a missing product milestone.

ichinya (changes requested):
break workflow

ichinya:
**Fix Checklist**

**Skill Contract**
- [ ] Add an explicit `Artifact Ownership / Config Policy` section to `/aif-qa`.
- [ ] State clearly that `/aif-qa` is `config-aware` but `read-only` for `config.yaml`.
- [ ] State clearly that `/aif-qa` must treat `DESCRIPTION.md`, `ARCHITECTURE.md`, `ROADMAP.md`, `RULES.md`, and `skill-context` as read-only inputs.
- [ ] State clearly that persistent writes are limited to `paths.qa/<branch>/`.
- [ ] Define the owned artifacts explicitly:
  - [ ] `change-summary.md`
  - [ ] `test-plan.md`
  - [ ] `test-cases.md`

**Workflow Correctness**
- [ ] Replace the branch-to-directory mapping with a filesystem-safe slug.
- [ ] Do not use `>` in branch-derived paths.
- [ ] Ensure branch names are safe on Windows, macOS, and Linux.
- [ ] Preserve the resolved branch across all stages.
- [ ] Make follow-up prompts include the resolved branch when suggesting the next command.
- [ ] Verify that `/aif-qa change-summary <branch>` -> `/aif-qa test-plan <branch>` works without re-resolving to the current checkout branch.
- [ ] Verify that `/aif-qa test-plan <branch>` -> `/aif-qa test-cases <branch>` works the same way.

**Tool Permissions**
- [ ] Fix the mismatch between instructions and allowed tools.
- [ ] If the workflow requires creating directories, allow the needed shell command explicitly.
- [ ] Or rewrite the workflow so it does not depend on a disallowed command.
- [ ] Re-check every command mentioned in the references against `allowed-tools`.

**Documentation Consistency**
- [ ] Update the workflow docs to include `/aif-qa`.
- [ ] Add `/aif-qa` to the artifact ownership table in the docs.
- [ ] Add `/aif-qa` to the config-aware skill lists everywhere they appear.
- [ ] Make sure `paths.qa` is documented consistently in:
  - [ ] config reference
  - [ ] configuration guide
  - [ ] skills reference
  - [ ] AGENTS/ownership docs
- [ ] Ensure docs consistently describe `/aif-qa` as a reader of config and a writer only to QA artifacts.

**Behavior Rules**
- [ ] Confirm that `/aif-qa` does not modify `config.yaml`.
- [ ] Confirm that `/aif-qa` does not modify roadmap, rules, architecture, or description artifacts.
- [ ] Confirm that `/aif-qa --all` still respects the same write boundaries.
- [ ] Confirm that failures stop the pipeline without partially advancing to the next stage.

**Validation / Tests**
- [ ] Add smoke coverage for the new skill.
- [ ] Add a test for branch slug sanitization.
- [ ] Add a test for explicit branch flow across all three stages.
- [ ] Add a test for default-path behavior when `config.yaml` is missing.
- [ ] Add a test proving the skill does not require undeclared tools.
- [ ] Add a test for `--all` mode end-to-end artifact generation.

**Merge Readiness**
- [ ] Re-run docs/integrity checks after updating the skill and references.
- [ ] Re-run skill validation after adding the ownership/config section.
- [ ] Confirm the final contract is: `read config, read context artifacts, write only under paths.qa/<branch>/`.

Velhron:
@ichinya Can you check it again? I kind of fixed it.

ichinya:
Rechecked on current PR head `a5734405235990daa3c9f7d2f7500d52c72fadb5`.

The earlier Windows-path, `mkdir` permission, and resolved-branch handoff issues look fixed. I still do not think this PR is merge-ready, though: there is one remaining artifact-isolation bug in the branch slugging scheme, one incomplete fallback in the base-branch flow, and one remaining discoverability/docs gap around the new built-in skill.

Context gates:
- Architecture: OK
- Rules: WARN — the new built-in skill is still not wired into all related surfaces, and I still do not see smoke coverage for slug collisions / explicit branch flow / `--all`.
- Roadmap: OK

Verification:
- `npm run build` passed
- `bash scripts/test-init.sh` passed
- `bash skills/aif-skill-generator/scripts/validate.sh skills/aif-qa` passed


## Comments :
* [P1] The branch -> artifact_dir mapping is still not guaranteed to be unique. This slugging scheme replaces `/` and other non-alphanumeric characters with `-`, so `feature/foo` and `feature-foo` both resolve to the same `feature-foo` directory. That means `change-summary.md` / `test-plan.md` / `test-cases.md` can overwrite each other across different branches, and later stages may read artifacts from the wrong branch. The artifact key needs to preserve uniqueness, not just filesystem safety.
* [P2] The “running QA on the base branch itself” fallback is only specified halfway. The instructions say to use `HEAD~1` when `resolved_branch == git.base_branch`, but the diff commands still compare `<base_branch>...<resolved_branch>`. As written, this can leave the skill with contradictory inputs: a fallback commit range for the log, but an empty diff range for the actual change analysis. The fallback needs to drive the diff range too, and it should be anchored to `resolved_branch`, not the current checkout’s `HEAD`.
* [P3] `aif-qa` is still missing from some user-facing discoverability surfaces. There is no human-readable hint for it in `src/cli/wizard/skill-hints.ts`, so the wizard will show it as “Additional custom skill”. `docs/workflow.md` still omits `/aif-qa` from the utility-command list, and `AGENTS.md` still says there are 22 built-in skills even though there are now 24.
* [P3] The docs still describe QA artifacts under `paths.qa/<branch>/...`, while the skill now writes under `<paths.qa>/<branch-slug>/...`. For branches like `feature/foo`, that sends users to the wrong directory and contradicts the skill contract. The docs should consistently say `branch-slug`, not `branch`.

Velhron:
@ichinya I do not know how to solve the P1 problem.
It seems to me that this is a rather rare and specific case. Of the options: 
1) We can check if there are artifacts in this folder and prompt the user to select another folder. But this will violate the `--all` mode.
2) Replace `/` with `--` — `feature/foo` → `feature--foo`, `feature-foo` → `feature-foo`. Flat structure, deterministic, but looks a bit unusual.
3) Preserve `/` as a directory separator — `feature/foo` → `qa/feature/foo/`. Uniqueness is structural, mkdir -p handles nested dirs naturally. Most readable.

What do you think?

ichinya:
Rechecked on current PR head `c516cdc`.

The earlier Windows-path, `mkdir` permission, resolved-branch handoff, base-branch fallback, wizard hint, and most of the discoverability/config wiring issues look fixed. I still do not think this PR is merge-ready, though: there is one remaining correctness issue in the branch -> artifact-dir mapping, plus a small set of stale docs around the final artifact-path contract.

Context gates:
- Architecture: OK
- Rules: WARN — I still do not see smoke coverage for branch-key uniqueness / explicit-branch flow / `--all`
- Roadmap: OK

Comments:
- [P1] The branch -> `artifact_dir` mapping is still lossy, so different branch names can still collapse onto the same QA directory. Under the current scheme, `feature/foo` and `feature-foo` both resolve to the same artifact location. That means one branch can overwrite another branch’s `change-summary.md` / `test-plan.md` / `test-cases.md`, and later stages may read the wrong artifacts. The artifact key needs to be injective, not just filesystem-safe.
- On the options in the thread: I would not ship option (2) as-is, because replacing `/` with `--` only fixes one collision class while the rest of the normalization remains lossy. Option (3) is acceptable only if the per-segment encoding is also lossless; preserving `/` as directories by itself does not solve collisions for other normalized characters. The safest fix is either a reversible encoding of the original branch name, or a readable slug plus a short hash suffix derived from the full branch name.
- [P3] Some docs still describe the QA path inconsistently. `docs/configuration.md`, `skills/aif/references/config-template.yaml`, and the Working Directory / Workflow Logic sections in `AGENTS.md` still use `<branch>` where the implementation now uses a derived artifact key. `docs/workflow.md` also still omits `/aif-qa` from the artifact ownership table. Please align all user-facing docs with the final path contract in the same patch.

Once the artifact key is collision-free and the remaining docs are aligned, I’m happy to recheck.

Velhron:
@ichinya Can you check it again?

ichinya:
Here is a detailed review that can be left in the PR.

In general, the PR is already noticeably closer to merge-ready: `/aif-qa` now has a clear `Artifact Ownership and Config Policy`, `allowed-tools` now covers both `Bash(git *)` and `Bash(mkdir *)`, and the handoff between stages actually preserves `<resolved_branch>` in follow-up commands for `change-summary -> test-plan -> test-cases`. Plus, a separate smoke-test `scripts/test-aif-qa.sh` has been added, and it is connected to the general `scripts/test-skills.sh`. 

However, I would still leave **Request changes** for now, mainly because of a single correctness issue in the branch slug contract, and there are a couple of smaller comments as well.

**1. [P2] The branch slug is not truly injective at the moment, although the PR promises otherwise.**
In `skills/aif-qa/SKILL.md`, the algorithm is described as `safe_slug + the first 8 hex characters of the git hash-object`, and it is explicitly stated that this makes the mapping injective and that different branch names “always resolve to different directories”. The same is repeated in the documentation, and the smoke-test specifically checks the wording about injective encoding. However, the 8 hex characters are not a reversible encoding of the branch name, but rather a short hash suffix; collisions for different branch names are still theoretically possible. That is, the contract is now stronger than the actual implementation. I would suggest one of two things: either make the mapping truly injective/reversible, or honestly rename the contract to something like “collision-resistant slug” and stop promising a guarantee that doesn’t actually exist. [aif-qa/SKILL.md](https://github.com/Velhron/ai-factory/blob/feature/aif-qa/skills/aif-qa/SKILL.md)

**2. [P3] The “Analyze only the last 20” option in `CHANGE-SUMMARY.md` does not currently limit the actual scope of the analysis.**
In reference, when there are >20 commits, the option “Analyze only the last 20” is initially suggested, but the diff is still constructed as `git diff <effective_base>...<resolved_branch>` for the entire branch. This means that the user chooses to reduce the scope, but the actual limitation is only on the commit list, not on the diff itself or the set of modified files. As a result, the UX is a bit deceiving and does not guarantee saving of context. I would either really shift the base for the “last 20” mode or explicitly rename the option so that it is clear: it is limited to commit log only, not the full change set. [aif-qa/references/CHANGE-SUMMARY.md](https://github.com/Velhron/ai-factory/blob/feature/aif-qa/skills/aif-qa/references/CHANGE-SUMMARY.md)

**3. [P3] The smoke-test on stage handoff is weaker than the regression that it is supposed to guard.**
`test-aif-qa.sh` currently only checks that the string `resolved_branch` appears somewhere in each reference file. This is not enough to fix the specific bug that was already caught in the review: it is possible to break the follow-up command again, but leave a single mention of `resolved_branch` in another part of the file, and the test will remain green. I would strengthen the smoke coverage to point checks of specific handoff-commands (`/aif-qa test-plan <resolved_branch>`, `/aif-qa test-cases <resolved_branch>`) and read/write checks specifically through `<artifact_dir>/...`. [scripts/test-aif-qa.sh](https://github.com/Velhron/ai-factory/blob/feature/aif-qa/scripts/test-aif-qa.sh)

**Result:**
The idea and structure of PR are good, and most of the previous blockers are already really closed. But before merging, I would still finish off point number 1: either make the slug contract really injective, or weaken the wording in `SKILL.md ` /docs /smoke-tests to an honest probabilistic/collision-resistant option. Items No.No. 2 and No. 3 no longer look blocking, but I would try to correct them in the same run. [aif-qa/SKILL.md](https://github.com/Velhron/ai-factory/blob/feature/aif-qa/skills/aif-qa/SKILL.md)

PR #70, opened 2026-04-09
feat(codex): materialize managed agent assets

pepperfm:
## Summary

- materialize Codex native agents and managed `.codex/config.toml` during init, update, and upgrade
- extend managed-state tracking for config files and harden managed path resolution
- move bundled Claude agents to an explicit `subagents/claude/agents/` source layout parallel to Codex
- reserve bundled runtime-owned agent files through builtin runtime metadata instead of hardcoded runtime ids
- document Codex agent support and add smoke coverage for drift, cleanup, round-trip, and Claude-only upgrade paths

## Behavioral Notes

- This PR adds baseline native-agent support for Codex, not full Claude parity. Codex gets the planning / implementation / review layer, while the broader Claude `loop-*` bundle remains Claude-only for now.
- AI Factory intentionally owns the generated `.codex/config.toml` in this PR. Drift is restored by `ai-factory update`; a future `--preserve-config` escape hatch is tracked as follow-up rather than implemented here.
- If a future package version drops a previously bundled source file, `ai-factory update` reports that managed agent file as skipped and preserves the local tracked file instead of deleting it implicitly. Deletion still belongs to explicit agent deselection or extension removal flows.
- Codex model identifiers were checked against local Codex CLI `0.128.0` via `codex debug models`; the catalog includes both `gpt-5.4` and `gpt-5.4-mini`, so the bundled TOML model strings are left unchanged.

## Verification

- `npm run build`
- `npm run lint`
- `npm test`
- `npm run test:init`
- `npm run test:update`
- `env npm_config_cache=/tmp/aif-npm-cache npm run test:extension-schema`
- `env npm_config_cache=/tmp/aif-npm-cache npm run test:extensions`


ichinya:
This is a good step for Codex support, but it is important to be clear that this is still partial parity with Claude, not full parity.

What looks good:
- Codex finally gets managed native agent assets instead of being skills-only: `.codex/agents/*.toml` plus a project-scoped `.codex/config.toml`.
- The installer/update/upgrade flow is wired through consistently, including drift healing and cleanup on deselection.
- The bundle split is done in the right direction: Claude stays on top-level `subagents/*.md`, while Codex gets its own runtime-specific bundle under `subagents/codex/...`, which avoids mixing Claude and Codex assets.
- The smoke coverage is strong for the intended scope: init, update, drift recovery, combined Claude+Codex installs, and Claude-only upgrade behavior.

Compared to the current Claude setup, though, Codex is still notably behind:
- Claude currently has the full subagent stack, including the entire `loop-*` family. Codex in this PR only gets the planning / implementation / review layer.
- The Claude agent contracts are much richer and more explicit about runtime behavior (top-level-only usage, delegation model, worktree/isolation expectations, merge/commit flow, Handoff sync semantics, etc.). The Codex TOML agents are much thinner and read more like a baseline adaptation than a real parity port.
- The docs now say AI Factory ships runtime-native assets for both Claude and Codex, but the current text can still imply a broader parity than the code actually provides.

What I’d like to see tightened before merge, or at least called out explicitly:
1. Document the non-goal clearly: this PR brings Codex to baseline native-agent support, not Claude-equivalent subagent parity.
2. Clarify ownership of `.codex/config.toml`. Right now update intentionally overwrites drift, which is fine only if we want AI Factory to fully own that file. If user customization is expected, that policy probably needs a narrower managed block or an explicit opt-out.
3. Bring the Codex coordinator/worker contracts closer to the Claude ones where possible, or explain why the reduced contract is intentional for Codex.
4. Track the missing `loop-*` parity as follow-up work if “Codex support” is meant to be comparable to Claude support rather than just bootstrap support.

Overall: I’m positive on the direction and the installer/state-management work looks solid. I would just avoid presenting this as “Claude and Codex are now equivalent” when the actual result is “Codex now has a managed native-agent baseline modeled after the Claude setup.”

ichinya (changes requested):
Further development is needed

ichinya:
## Main remarks:
- `[P0]` The PR branch is not being collected. In [installer.ts](<src/core/installer.ts:1104>), the closing `}` is lost. The local `tsc --noEmit' crashes.
- `[P1]` In [init.ts](<src/cli/commands/init.ts:102>), cleanup is broken when deselect is applied to extension-defined runtimes. The logic is now tied to the bundled-source resolver, which doesn't work for extension runtime.
- `[P2]` In [scripts/test-update.sh](<scripts/test-update.sh:245>), the tests don't match the actual CLI output: they expect `Subagents`, but the code prints `Agent files`.
- The PR has explicit merge leftovers: duplicate imports in [upgrade.ts](<src/cli/commands/upgrade.ts:15>) and [installer.ts](<src/core/installer.ts:22>). These should be removed before rebuilding.
- Scope PR should be kept on top of merged [#71](https://github.com/lee-to/ai-factory/pull/71): avoid confusing the description and terminology around the general `agentFiles` infrastructure, and focus this PR on the Codex-specific delta.

## What's good in the PR:
- Added bundled Codex native agents in `subagents/codex/agents/*.toml` and managed `.codex/config.toml`.
- `init/update/upgrade` are extended consistently, including managed state and drift healing.
- Codex contracts are now handoff-aware, which is a useful approximation of the Claude flow.
- The documentation better reflects the actual scope: baseline support for Codex, rather than parity with Claude.
- The smoke coverage for Codex scenarios is good as intended: init, update, drift, combined install, and Claude-only upgrade.

## What I would remove or redo:
- Fix build-breaking errors first, and only then discuss UX and wording.
- Cleanup for removed runtime so that it resolves the target path regardless of whether it is a built-in runtime or an extension runtime.
- Choose one terminology and use it everywhere. After #71, it makes more sense to stick to the universal `Agent files` rather than mixing `subagents` and `agent files`.
- Minimize merge leftovers by removing duplicate imports, inconsistent test strings, and any traces of an incomplete merge.
- Keep the PR and docs description as a follow-up to `#71`: “bundled Codex assets + managed config + handoff-aware contracts”, without re-expanding the general narrative about the extension runtime platform.

ichinya (changes requested):
The PR direction is correct, but it is not currently merge-ready

ichinya:
Overall this is much closer after the follow-up fixes, and the scope/docs are now in the right place. I still think this should stay in Request changes for two blockers:

1. [P1] Bundled agent-file ownership conflicts are still Claude-only.
`assertNoAgentFileConflicts()` still reserves only bundled Claude files via `getAvailableSubagents()`, but this PR now adds bundled Codex agent files under `subagents/codex/agents/*.toml`. That leaves a hole where an extension can still target the same Codex file path without tripping the ownership guard. Please reserve bundled Codex files in the same ownership map and add a smoke case for a Codex collision.

2. [P1] The read-only Codex sidecars do not encode the read-only contract in config.
`implement-coordinator` is explicitly told to delegate read-only audits to `review-sidecar`, `security-sidecar`, `best-practices-sidecar`, `docs-auditor`, and `commit-preparer`, but those read-only sidecars/auditors omit `sandbox_mode`. In Codex custom-agent files, omitted `sandbox_mode` inherits from the parent session, so these roles are currently read-only only by prompt text. Please add `sandbox_mode = "read-only"` to the read-only sidecars/auditors so the contract is reflected in the agent config as well.

After those two are fixed, I’m good with the direction. Bundled Codex agents plus a managed `.codex/config.toml`, along with the init/update/upgrade plumbing and the clarified docs scope, all make sense for baseline Codex support.

pepperfm:
@ichinya Закрыл два последних блокера в этой ветке.

Что поменял:
- bundled Codex agent files теперь участвуют в ownership-conflict guard так же, как bundled Claude files, поэтому extension больше не может занять bundled Codex target вроде `review-sidecar.toml`
- у read-only Codex sidecars/auditors теперь явно задан `sandbox_mode = "read-only"` в agent config, а не только read-only формулировка в prompt
- в `scripts/test-init.sh` добавил smoke coverage и на bundled Codex collision, и на явный read-only sandbox contract у установленных Codex sidecars

Что прогнал локально:
- `npm run test:init`
- `npm run test:update`

Если посмотришь PR ещё раз, это должно закрывать два блокера из последнего комментария.

ichinya:
Thanks, the two previous blockers are addressed now:
- bundled Codex agent files are included in the ownership map alongside Claude;
- the read-only Codex sidecars/auditors now declare `sandbox_mode = "read-only"`, and the smoke coverage checks that contract.

I still found one ownership blocker before merge:

[P1] Normalize or reject non-canonical extension agent-file targets before conflict checks.

`assertNoAgentFileConflicts()` compares ownership using the raw manifest `agentFile.target`, while installation later resolves that target as a filesystem path under the runtime `agentsDir`. That means an extension can bypass the bundled-file ownership guard by using a path alias such as `./review-sidecar.toml` or `nested/../review-sidecar.toml`: the ownership key no longer equals `codex::review-sidecar.toml`, but the resolved target on disk is still `.codex/agents/review-sidecar.toml`.

Please canonicalize the target once and use that canonical value consistently for:
- ownership conflict checks;
- installedAgentFiles / agentFileSources state;
- install/remove/update paths;

or reject non-canonical targets outright when `path.posix.normalize(target) !== target`.

Please also add a smoke case for a bundled Codex collision using a non-canonical target, for example `target: "./review-sidecar.toml"`.

After that, I’m good with the PR. The Codex bundle, managed `.codex/config.toml`, drift recovery, deselect cleanup, and the clarified baseline-vs-parity docs all look aligned with the intended scope.

pepperfm:
@ichinya Поднял ветку на свежий upstream 2.x и дотащил follow-up fix по последнему blocker.

Что в актуальном состоянии ветки:
- merge с upstream 2.x
- reject non-canonical agentFiles.target (например ./review-sidecar.toml) на этапе manifest validation
- smoke case на non-canonical Codex collision
- локально прогнаны npm run test:init и npm run test:update

Если посмотришь ещё раз, нужен уже только re-check по текущему состоянию ветки после merge и этого fix.

ichinya:
Review: Changes requested.

The previous blockers are addressed: bundled Codex agent files are now in the ownership map, and read-only sidecars/auditors encode sandbox_mode="read-only" with smoke coverage.

One ownership blocker remains before merge:
Please normalize or reject non-canonical extension agent-file targets before conflict checks. Comparing raw manifest target strings allows aliases like `./review-sidecar.toml` or `nested/../review-sidecar.toml` to bypass the ownership guard while still resolving to `.codex/agents/review-sidecar.toml`.

Expected fix:
- canonicalize target paths under the destination agentsDir before conflict detection;
- reject paths that escape the destination;
- use the canonical destination as the ownership key;
- add smoke tests for `./review-sidecar.toml`, `nested/../review-sidecar.toml`, and path escape attempts.

After that, this PR should unblock the Handoff native Codex path.

pepperfm:
@ichinya

ichinya:
Thanks, this is much closer. The previous blockers are addressed:
- bundled Codex agent files are now reserved alongside Claude;
- the read-only Codex sidecars/auditors now encode `sandbox_mode = "read-only"`;
- the smoke coverage now includes the Codex alias/path-escape cases.

I still think this needs one backward-compat fix before merge:

[P2] Do not hard-fail legacy installed extension manifests that used previously accepted agent-file target aliases.

The new manifest validation correctly rejects non-canonical `agentFiles.target` values such as `./review-sidecar.toml` and `nested/../review-sidecar.toml`. The problem is that `loadExtensionManifest()` now applies that validation when reading already-installed manifests, and `loadAllExtensions()` does not catch or skip those validation errors. In practice, that means a project with a pre-fix installed extension using an alias target can now fail `extension add` / `extension update` until the stored extension is manually cleaned up.

That is also inconsistent with the existing update UX: ordinary `ai-factory update` already treats missing/invalid installed extension manifests as recoverable and preserves tracked agent-file state instead of treating the project as unrecoverable.

Please either:
- normalize legacy installed `agentFiles.target` values on read before validation / conflict reconstruction; or
- treat invalid already-installed manifests the same way ordinary update does (warn + skip for ownership reconstruction), while keeping the stricter validation for new installs and updates.

Please also add one smoke case for a previously installed extension manifest with `target: "./review-sidecar.toml"` and verify that a subsequent `extension add` or `extension update` still succeeds with a clear warning or normalized handling.

After that, I’m good with the PR. The runtime-specific Codex bundle, managed `.codex/config.toml`, drift healing, deselect cleanup, and the clarified baseline-vs-parity docs all look aligned with the intended scope.

ichinya:
Looks good to me now.

The remaining backward-compat blocker from my Apr 29 review looks addressed. Installed extension manifests now go through a tolerant read path that normalizes safe legacy agent-file target aliases before validation, and the installed-manifest consumers were switched over accordingly. The new smoke coverage also exercises the actual regression path: mutate a previously installed manifest to `./legacy-helper.toml`, then verify both `extension add` and `extension update --force` still proceed with a warning and that update rewrites the stored manifest back to the canonical target.

I don’t see a remaining merge blocker in the code, so approving from the review side. I’d still treat repo-side checks separately, since GitHub currently isn’t showing runnable jobs under Checks.

ichinya (approved):

lee-to:
# Review: feat(codex): materialize managed agent assets

Solid extension of managed-asset pattern. Core shape correct. Concerns below.

## High

**`resolveManagedConfigFilePaths` skips boundary check** — `src/core/installer.ts:285-302`. Calls `assertSafeManagedRelativePath` only. Sibling `resolveInstalledAgentFileTargetPath` also calls `ensureTargetWithinRoot`. Asymmetry. `assertSafeManagedRelativePath` rejects `/abs`, `..`, `../x` but lets Windows-drive-style `C:/foo` through (post `replaceAll('\\','/')`). Add `ensureTargetWithinRoot` for both `targetFile` vs `configDir` root and `sourceFile` vs `sourceRoot`. Not theoretical when `legacyAgent.configFiles` round-trips from hand-edited `.ai-factory.json`.

**`getAvailableSubagents` claude filter relies on tree-layout convention** — `src/core/installer.ts:419-428`. `subagents/` now houses flat claude `.md` files and nested `codex/agents/*.toml`. Filter `!relPath.includes('/')` works today but silently breaks the moment a subdir of claude assets lands under `subagents/`. Structural invariant. Consider moving claude bundle to `subagents/claude/agents/` with explicit `agentsSourceDir` parallel to codex — filter logic disappears.

## Medium

**Hardcoded runtime list in `assertNoAgentFileConflicts`** — `src/core/extension-ops.ts:361-373`. `bundledRuntimeIds = [AGENT_IDS.claude, AGENT_IDS.codex]` plus per-id label switch. Better: iterate `BUILTIN_AGENT_REGISTRY` entries where `agentsSourceDir` set, derive label from `displayName`. Avoids touching this for every new bundled runtime.

**`updateSubagents` source-missing semantics flipped** — `src/core/installer.ts:1098`. Was `install: true`, now `install: false` with status `skipped`. Behavior change when package later drops a bundle file. Intentional, but worth one-line note in PR body.

**`.codex/config.toml` drift overwritten with only `console.warn`** — `src/core/installer.ts:1257`. Mirrors skills/subagents pattern, but `config.toml` is runtime config users may legitimately tune (`max_threads`, `job_max_runtime_seconds`). Doc covers intent. Follow-up: `--preserve-config` escape hatch. Not blocking.

## Low

**Dead fallback in `getBundledAgentFilesSourceDir`** — `src/core/installer.ts:308-318`. `agentId === AGENT_IDS.claude ? getSubagentsDir() : null` unreachable since claude has explicit `agentsSourceDir: 'subagents'`. Drop fallback.

**Model strings `gpt-5.4` / `gpt-5.4-mini`** — TOMLs across `subagents/codex/agents/*.toml`. Verify these resolve as Codex CLI model identifiers — if not, agents fail at runtime.

**`AGENT_IDS` const placement** — `src/core/agents.ts:34-37`. Inserted between `RuntimeDefinitionInput` and `RuntimeManifestInput` interfaces. Move above interfaces or into constants block.

**Style nit** — `src/core/installer.ts:303` missing blank line before `function getBundledAgentFilesSourceDir`.

## Good

- Legacy agent-file target alias tolerance (`loadInstalledExtensionManifest`) is the right migration shape — strict for new validations, lenient for installed manifests with warn log.
- `removeAgentSetup` now consumes installed extension manifests to clean orphaned extension agent files on agent deselect (`src/cli/commands/init.ts:104-116`). Closes real gap.
- `assertSafeManagedRelativePath` centralized. `resolveManagedSubagentPaths` removes duplicated pairs.
- Smoke tests cover codex init, claude+codex coexistence, drift recovery, cleanup, claude-only upgrade.
- Docs honest about scope: "baseline ... not full parity".

## Verdict

Approve after High items addressed (boundary check on config-file resolution, claude-bundle layout decision). Medium items are follow-ups.

lee-to:
Thanks for the fixes. I re-reviewed the current head and found two remaining issues:

1. `ai-factory update` silently overwrites an existing user-owned `.codex/config.toml` during migration.

For old Codex configs without `configFiles`, `loadConfig()` backfills `configFiles` from the runtime default, then `updateConfigFiles()` treats `config.toml` as `new-in-package` and copies the bundled file over whatever already exists at `.codex/config.toml`.

I reproduced this with a project that already had:

```toml
[user]
keep = true
```

After `ai-factory update`, the file became:

```toml
[agents]
max_threads = 6
max_depth = 2
job_max_runtime_seconds = 1800
```

This should either skip/adopt-with-warning when an untracked file already exists, or write the managed config to a non-conflicting path. Relevant areas: `src/core/config.ts` around the `configFiles` backfill, and `src/core/installer.ts` in `updateConfigFiles()`.

2. Unsafe `installedConfigFiles` paths abort agent deselection instead of being skipped like unsafe agent-file paths.

`removeAgentSetup()` catches unsafe `installedAgentFiles`, but then removes `installedConfigFiles` without the same guard. A config containing `installedConfigFiles: ["../SHOULD_NOT_DELETE.toml"]` makes `init --agents claude` crash while deselecting Codex. It does not delete outside the config dir, but it leaves the project unreconfigurable until the JSON is manually edited. This should mirror the warning-and-skip behavior already used for managed agent files. Relevant area: `src/cli/commands/init.ts` around the config file removal loop.

Verification: `npm test` reached the gate-result smoke test and then failed with an apparently unrelated shell issue:

```text
scripts/test-gate-result-contract.sh: line 183: verify: unbound variable
```

The init/update smoke tests passed before that failure, and I verified the two scenarios above manually.

lee-to:
Thanks for the follow-up. I still see two overwrite paths for user-owned Codex config that are not covered by the current fix:

1. `init --agents codex` still overwrites a pre-existing `.codex/config.toml`.

`initCommand()` calls `installConfigFiles()` directly, and `installConfigFiles()` unconditionally copies the bundled `subagents/codex/config.toml` to the project target. Repro: create `.codex/config.toml` with user content, then run `init --agents codex --skills aif`; the file is replaced with the managed `[agents]` defaults. This means the migration protection added in `updateConfigFiles()` does not protect first-run/re-run init.

Relevant paths: `src/cli/commands/init.ts`, `src/core/installer.ts`.

2. `update --force` also overwrites an untracked pre-existing `.codex/config.toml`.

In `updateConfigFiles()`, the `force` branch runs before the untracked-target check, so a legacy project with no `installedConfigFiles` state but an existing `.codex/config.toml` gets force-reinstalled instead of preserved. Repro: create a legacy `.ai-factory.json` for Codex, add user content to `.codex/config.toml`, then run `ai-factory update --force`; the file is replaced with the managed defaults.

Relevant path: `src/core/installer.ts`.

I think both paths need the same ownership rule as normal update: if the target exists but is not tracked as installed/managed by AI Factory, preserve it and do not add it to `installedConfigFiles` / `managedConfigFiles`.

PR #71, opened 2026-04-11
Add extension-managed agent files for Codex and dynamic runtimes

ichinya:
This PR adds first-class support for runtime-aware agent files in AI Factory extensions.

Before this change, AI Factory could install bundled Claude subagents from the package, but extensions could not install custom agent assets into runtime-specific agent directories such as `.codex/agents/` or extension-defined agent folders. Extension manifests could declare `agents`, but those runtime definitions were not fully integrated into the active registry or install lifecycle.

This change introduces a new manifest field, `agentFiles`, separate from `agents`.

What this enables:
- Extensions can now install custom Claude agent files into `.claude/agents/`
- Extensions can now install Codex custom agents into `.codex/agents/*.toml`
- Extensions can provide their own runtimes and install agent files into those runtime-specific `agentsDir` locations
- `init`, `update`, `upgrade`, `extension add`, `extension update`, and `extension remove` now participate in the same managed lifecycle for these files

Key changes:
- Added universal agent-file state in `.ai-factory.json`:
  - `agentsDir`
  - `installedAgentFiles`
  - `managedAgentFiles`
- Kept backward compatibility by reading legacy Claude-only keys:
  - `subagentsDir`
  - `installedSubagents`
  - `managedSubagents`
- Hydrated the effective runtime registry from installed extensions before CLI validation and runtime lookups
- Added `agentFiles` manifest validation:
  - runtime must exist
  - source and target must be safe relative paths
  - file extensions must match the runtime format
  - ownership conflicts on `runtime + target` are rejected before install
- Unified bundled Claude agent files and extension-provided agent files under one managed artifact pipeline
- Added rollback and orphan-runtime protection for extension update/remove flows

Why this matters:
- Codex users can now receive custom agents from extensions without manual copying
- Extension authors can ship runtime-specific agent assets in a predictable, managed way
- AI Factory now cleanly separates two concepts:
  - `agents`: define runtimes
  - `agentFiles`: provide agent assets for those runtimes

Also included:
- Documentation updates for extensions, configuration, getting started, and Claude subagents
- Smoke test coverage for:
  - Codex agent files
  - dynamic runtime registration
  - extension lifecycle install/update/remove
  - legacy config migration
  - conflict and orphan-runtime protections
- Updated example extension showing `agentFiles` for Claude, Codex, and an extension-defined runtime

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,9 +11,9 @@
     {
       "id": "claude",
       "skillsDir": ".claude/skills",
-      "subagentsDir": ".claude/agents",
+      "agentsDir": ".claude/agents",
       "installedSkills": ["aif", "aif-plan", "aif-improve", "aif-implement", "aif-commit", "aif-build-automation"],
-      "installedSubagents": [
+      "installedAgentFiles": [
         "best-practices-sidecar.md",
         "commit-preparer.md",
         "docs-auditor.md",
@@ -44,6 +44,7 @@
     {
       "id": "codex",
       "skillsDir": ".codex/skills",
+      "agentsDir": ".codex/agents",
       "installedSkills": ["aif", "aif-plan", "aif-implement"],
       "mcp": {
         "github": false,
@@ -64,7 +65,7 @@
 }
 ```
 
-The `agents` array can include any supported agent IDs: `claude`, `cursor`, `windsurf`, `roocode`, `kilocode`, `antigravity`, `opencode`, `warp`, `zencoder`, `codex`, `copilot`, `gemini`, `junie`, or `universal`. Each agent keeps its own `skillsDir`, installed skills list, and MCP preferences. Claude Code agents also persist `subagentsDir` and `installedSubagents`, so `ai-factory update` can refresh `.claude/agents/` alongside skills. AI Factory additionally stores internal `managedSkills` and `managedSubagents` hash maps in `.ai-factory.json`; they are omitted from the example above for brevity.
+The `agents` array can include any built-in agent IDs plus runtime IDs provided by installed extensions. Each agent keeps its own `skillsDir`, installed skills list, and MCP preferences. Runtimes that support custom agent files also persist `agentsDir` and `installedAgentFiles`, so `ai-factory update` can refresh package-managed agent files alongside skills. AI Factory additionally stores internal `managedSkills` and `managedAgentFiles` hash maps in `.ai-factory.json`; they are omitted from the example above for brevity. `loadConfig()` still reads legacy Claude-only `subagentsDir`, `installedSubagents`, and `managedSubagents` keys for backward compatibility, but new saves use the universal field names.
 
 The optional `extensions` array tracks installed extensions by name, original source, and version. `ai-factory update` now refreshes these extensions from their saved sources before base-skill updates, and `ai-factory extension update [name] --force` refreshes them without running the full base-skill update flow.
 
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -2,7 +2,7 @@
 
 # Extensions
 
-Extensions let third-party developers add new capabilities to AI Factory — custom CLI commands, MCP servers, skill injections, agent definitions, and more. Extensions survive `ai-factory update` (injections are automatically re-applied after skills are refreshed).
+Extensions let third-party developers add new capabilities to AI Factory — custom CLI commands, MCP servers, skill injections, runtime definitions, runtime-specific agent files, and more. Extensions survive `ai-factory update` (injections and managed agent files are automatically re-applied after skills are refreshed).
 
 ## For Users
 
@@ -43,9 +43,11 @@ ai-factory extension remove aif-ext-example
 1. Extension files are copied to `.ai-factory/extensions/<name>/`
 2. Extension is recorded in `.ai-factory.json` under `extensions`
 3. Extension skills (from `skills`) are installed into configured agents
-4. Injections are applied to matching skill files (e.g. appending extra instructions to `/aif-implement`)
-5. MCP servers are merged into each agent's settings file (e.g. `.mcp.json`)
-6. Custom CLI commands become available immediately
+4. Runtime definitions from `agents` are added to the effective registry for future `ai-factory init` runs
+5. Agent files from `agentFiles` are copied into matching runtime `agentsDir` locations
+6. Injections are applied to matching skill files (e.g. appending extra instructions to `/aif-implement`)
+7. MCP servers are merged into each agent's settings file (e.g. `.mcp.json`)
+8. Custom CLI commands become available immediately
 
 ### What Happens on Update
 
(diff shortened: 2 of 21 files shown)


lee-to:
## Code Review

**Verdict: Request changes** (minor, not blocking)

Design is sound — cleanly separates `agents` (runtimes) from `agentFiles` (managed assets), with a proper one-shot `subagentsDir → agentsDir` migration in `loadConfig` and validated install/rollback paths.

### Top concerns

1. **Swallowed copy failures** — `src/core/installer.ts` (`installExtensionAgentFiles`): failed agent-file copies only `console.warn` while the surrounding extension install reports success. Skill/MCP installs throw and trigger `rollbackFailedExtensionInstall`; agent files should too, otherwise a partial payload is persisted as "successful".

2. **`init` skips conflict validation** — `src/cli/commands/init.ts` (re-apply loop around `installExtensionAgentFilesForAllAgents`): re-running `init` does not call `assertNoAgentFileConflicts`, so an extension can silently overwrite a bundled Claude file that `extension add` would have rejected.

3. **No legacy-migration test** — `scripts/test-update.sh` asserts the new `agentsDir`/`installedAgentFiles` keys but nothing seeds a legacy `subagentsDir` config and verifies the read-then-migrate path end-to-end. Users upgrading from 2.x will exercise this blindly; please add a fixture test.

4. **Bundled ownership uses source filenames** — `src/core/extension-ops.ts` (ownership map seeded from `getAvailableSubagents()`): this reserves every shipped filename permanently, so an extension can never share a filename with any bundled file even if the user unmanaged it. Acceptable as defense-in-depth, but deserves a line in `docs/extensions.md` under the rules list.

5. **Eager hydration per CLI call** — `src/cli/index.ts` (`bootstrapProjectAgentRegistry()` awaited before Commander parses): every `ai-factory …` invocation walks `.ai-factory/extensions/<name>/extension.json` for all recorded extensions, even for `--help`/`--version`. Consider deferring into the individual command action handlers.

### Nits

- `src/core/extensions.ts` — `.replaceAll('\\','/')` + `path.isAbsolute` won't catch Windows `C:\…` drives in manifests authored on Windows.
- `ensureTargetWithinRoot` uses `path.resolve` which doesn't follow symlinks; a symlinked `.claude/agents → /tmp/evil` would not be detected. Not a regression, worth a comment.
- `getManifestRuntimeIds` / `assertNoConfiguredRuntimeOrphans` are duplicated between `src/cli/commands/extension.ts` and `src/core/extension-ops.ts` — pull into one shared helper.
- `scripts/test-init.sh` covers orphan protection and the happy path but never exercises the ownership-conflict rejection (two extensions claiming the same `claude::…`, or an extension colliding with a bundled Claude file). Add a negative case.
- `UpdateSubagentsResult.installedSubagents` — rename to `installedAgentFiles` for consistency with the new state schema (`src/cli/commands/update.ts`).
- `registerRuntimeDefinitions` duplicate-id branch is effectively dead code after `resetExtensionAgentRegistry()` in `hydrateProjectAgentRegistry`.

### Summary

Solid work. Rollback and orphan-runtime protection are implemented at the right points (`commitResolvedExtension` gates FS writes behind both `assertNoConfiguredRuntimeOrphans` and `assertNoAgentFileConflicts`), and path-traversal checks (`isSafeRelativeAssetPath` + `ensureTargetWithinRoot`) are defense-in-depth. The two items worth fixing before merge are (1) swallowed copy errors and (2) init's unvalidated re-apply path — both small surgical changes. Everything else is polish and test coverage.

ichinya:
Thanks for the detailed review. I went through each point and prepared follow-up changes for the requested fixes.

Addressed items:
- extension agent-file installs now fail fast, so copy failures abort the install and flow into the existing rollback path instead of being logged and treated as success
- `init` now runs the same `agentFiles` conflict validation before re-applying extension assets, so it can no longer bypass the checks enforced by `extension add` / `extension update`
- added legacy migration coverage for `subagentsDir` / `installedSubagents` / `managedSubagents` -> `agentsDir` / `installedAgentFiles` / `managedAgentFiles`
- documented that bundled Claude target filenames are reserved by `runtime + target`
- added a negative smoke test for bundled Claude ownership conflicts

I also cleaned up the related nits while touching the area:
- removed eager registry hydration from CLI bootstrap
- hardened manifest path validation for Windows-style absolute paths
- rejected symlinked `agentsDir` roots in managed agent-file installs
- deduplicated the orphan-runtime helper and the extension CLI config/hydration bootstrap
- renamed the update result field to `installedAgentFiles` for consistency
- clarified why the duplicate runtime-id guard still exists in `registerRuntimeDefinitions`

Local verification:
- `npm run build`
- `npm run lint`
- `npm run test:init`
- `npm run test:update`
- `npm test`

lee-to:
## Re-review after fix commit `9f2afa4`

All five top concerns from the previous review are addressed cleanly:

1. **Swallowed copy failures** — `src/core/installer.ts`: `installExtensionAgentFiles` now throws the new `AgentFileInstallError` carrying `installedTargets`, and `installExtensionAgentFilesForAllAgents` re-throws so the existing rollback path actually runs. 

2. **`init` missing conflict validation** — `src/cli/commands/init.ts`: the re-apply loop now calls `assertNoAgentFileConflicts` before `installExtensionAgentFilesForAllAgents`, so an extension colliding with a bundled Claude file can no longer silently overwrite it on `init`. 

3. **Legacy-migration test** — `scripts/test-update.sh`: new smoke test seeds an `.ai-factory.json` with legacy `subagentsDir` / `installedSubagents` / `managedSubagents`, runs `update`, and asserts via `node -e` that the legacy keys are gone and `agentsDir` / `installedAgentFiles` / `managedAgentFiles` are persisted. Exactly the coverage that was missing. 

4. **Reserved-filenames docs note** — `docs/extensions.md`: added a rule bullet clarifying that bundled Claude filenames are reserved by target path from the package `subagents/` inventory. 

5. **Eager hydration per CLI call** — `src/cli/index.ts`: `bootstrapProjectAgentRegistry()` removed entirely. Hydration is now deferred into the new `loadHydratedExtensionConfig` helper in `src/cli/commands/extension.ts`, called per command action. Bonus: four copies of the "loadConfig + hydrate + error" boilerplate collapsed into one helper. 

### Nits — also fixed

- `isSafeRelativeAssetPath` (`src/core/extensions.ts`) now rejects Windows drive-letter paths (`C:\…`) and UNC paths (`\\server\…` / `//server/…`).
- `ensureTargetWithinRoot` (`src/core/installer.ts`) rejects a symlinked root via `existsSync` + `lstatSync`, closing the `.claude/agents → /tmp/evil` redirection vector.
- `getManifestRuntimeIds` and `assertNoConfiguredRuntimeOrphans` are exported from `src/core/extension-ops.ts` and the duplicates in `src/cli/commands/extension.ts` are gone.
- `scripts/test-init.sh` now includes a negative conflict case: an extension trying to claim `claude/plan-polisher.md` is rejected by `extension add` with the expected `already owned by AI Factory bundled Claude agent files` message.
- `UpdateSubagentsResult.installedSubagents` → `installedAgentFiles` renamed consistently across `installer.ts` and `update.ts`.
- `registerRuntimeDefinitions` duplicate-id branch kept, now with a comment explaining it still guards collisions between different installed extensions and manifests validated in the same hydration pass. Reasonable.

### Minor observation (non-blocking)

In `src/core/extension-ops.ts` around the new catch block in `installExtensionAgentFilesForAllAgents`:

```ts
} catch (error) {
  if (error instanceof AgentFileInstallError) {
    results.set(agent.id, error.installedTargets);
  }
  throw error;
}
```

The `results.set(...)` right before `throw error` is effectively dead — `results` is discarded when the function unwinds, and the partial state is actually propagated via `AgentFileInstallError.installedTargets` on the error itself. Harmless, but the line can be removed (or replaced with a comment) to avoid suggesting that `results` is observable on the error path.

### Verdict

Good to merge. All previously flagged issues are addressed, the new negative tests cover paths that were previously uncovered, and the backward-compat migration is now actually exercised end-to-end. Nice turnaround.

ichinya:
Follow-up cleanup is pushed.

I removed the dead `results.set(agent.id, error.installedTargets)` line from the error path in `installExtensionAgentFilesForAllAgents`, along with the now-unused `AgentFileInstallError` import from that file.

Verification after the cleanup:
- `npm run build`
- `npm run lint`

This should close the last non-blocking observation from the re-review.

PR #75, opened 2026-04-14
feat(implement-coordinator): skip review/security sidecars when HANDOFF_SKIP_REVIEW=1

lee-to:
## Summary
- In single-task execution, skip `review-sidecar` and `security-sidecar` when `HANDOFF_SKIP_REVIEW=1`
- `best-practices-sidecar`, `docs-auditor`, `commit-preparer` continue to run
- Rationale: in autonomous Handoff mode with review skipped, there is no human to act on sidecar findings

## Test plan
- [ ] Run implement-coordinator with `HANDOFF_SKIP_REVIEW=1` — verify review/security sidecars are not spawned
- [ ] Run without the flag — verify all sidecars still spawn as before

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/subagents/implement-coordinator.md
+++ b/subagents/implement-coordinator.md
@@ -159,8 +159,8 @@ Workflow for single-task execution:
 2. Implement the target task using direct tool calls (Read, Write, Edit, Glob, Grep, Bash).
 3. Run one `aif-verify`-compatible verification pass scoped to the changed files.
 4. Launch read-only quality sidecars in background on the changed scope:
-    - `review-sidecar` — correctness, regression, performance risks
-    - `security-sidecar` — security audit
+    - `review-sidecar` — correctness, regression, performance risks — **skip if `HANDOFF_SKIP_REVIEW=1`**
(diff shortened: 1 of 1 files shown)


ichinya:
Looks good to me overall.

This change is small, but the intent is clear and the behavior makes sense: when `HANDOFF_SKIP_REVIEW=1`, spawning `review-sidecar` and `security-sidecar` adds cost without producing actionable output for an autonomous handoff flow, while keeping `best-practices-sidecar`, `docs-auditor`, and `commit-preparer` still preserves useful quality gates.

I also checked the diff itself: this PR only updates `subagents/implement-coordinator.md`, so this is really a subagent contract / behavior-spec change rather than a runtime implementation change. That is fine if this file is the source of truth for the bundled Claude subagent behavior.

Minor nit: the PR title `feat(implement-coordinator)` can read like a code/runtime change, while the actual diff is spec/prompt-level only. Not a blocker, but it may be worth reflecting that a bit more explicitly in the title or description for future readers.

From the workflow perspective, the change is coherent and reduces unnecessary sidecar work in skip-review mode. I’m good with this.

ichinya (approved):
LGMD

PR #79, opened 2026-04-15
fix(aif): resolve localization before setup artifacts

### LINKED ISSUE #77: [Bug] `/aif`: Inconsistent artifact localization due to late language selection
## Description:  
The language selection prompt occurs too late in the workflow, leading to a "mixed-language" state in generated artifacts.

## Current Behavior:
1. The `/aif` skill starts generating artifacts (usually in English by default).
2. Somewhere in the middle of the process, the agent asks the user for their preferred language.
3. As a result, some artifacts remain in English while others are created in Russian, or the language setting is ignored entirely, and all artifacts are generated in English.

## Expected Behavior:  
The language preference should be established at the **very beginning** of the session, before any artifacts are generated. This ensures consistency across all artifacts created by the skill.

ichinya:
Align /aif setup flow with early language resolution and ensure setup-time artifacts use the resolved artifact language consistently.

- resolve language.ui and language.artifacts immediately after mode detection
- write config.yaml before the first artifact and before /aif-architecture
- localize DESCRIPTION.md, rules/base.md, AGENTS.md, and UI summary templates
- add regression checks for localization order and hard-coded English leaks

Fixes #77

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -30,7 +30,9 @@ During setup, `/aif` resolves `language.ui` and `language.artifacts` immediately
 
 - If both language keys already exist in `config.yaml`, `/aif` reuses them and does not ask again.
 - If only one language key exists, `/aif` keeps the existing value and resolves only the missing key via `config.yaml` → `AGENTS.md` → `CLAUDE.md` → `RULES.md` → user question.
-- After language resolution, `/aif` refreshes `config.yaml` from `skills/aif/references/config-template.yaml` before writing `.ai-factory/DESCRIPTION.md`, `.ai-factory/rules/base.md`, `AGENTS.md`, or invoking `/aif-architecture`.
+- When `.ai-factory/config.yaml` already exists, `/aif` merges the resolved `language.ui` / `language.artifacts` values into the existing file and preserves other sections such as `paths.*`, `git.*`, `workflow.*`, and `rules.*`.
+- `/aif` preserves an existing `language.technical_terms` value and defaults it to `keep` only when the key is missing.
+- After language resolution, `/aif` updates `config.yaml` from `skills/aif/references/config-template.yaml` before writing `.ai-factory/DESCRIPTION.md`, `.ai-factory/rules/base.md`, `AGENTS.md`, or invoking `/aif-architecture`.
 - This ordering keeps all setup-time artifacts in a single run aligned to one `language.artifacts` value, while prompts, questions, summaries, and next-step guidance use `language.ui`.
 
 ## Schema Summary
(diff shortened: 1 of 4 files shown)


lee-to:
## Review notes

Overall the direction is right — resolving language before any setup artifact and persisting `config.yaml` up front is the correct fix for #77. A few concerns worth addressing before merge:

1. **Risk of overwriting user settings in `config.yaml` (Mode 1)** — `skills/aif/SKILL.md` Step 7 says to write `config.yaml` from `skills/aif/references/config-template.yaml`. If the file already exists with custom `paths.*`, `git.*`, or `rules.*`, a full rewrite from template will clobber them. The "Persist resolved settings" section mentions preserving inline comments but does not explicitly require merging with existing non-language keys. Please state explicitly: if `config.yaml` already exists, merge — only the resolved `language.*` keys may be replaced; other sections must be preserved.

2. **Brittle regression checks in `scripts/test-skills.sh`** — the new grep guards use fixed strings like `## Overview`, `## Tech Stack`, `| File | Purpose |`, `## Core Features`. These will false-positive on any unrelated mention anywhere in `skills/aif/SKILL.md` (examples, prose, future edits). The localized-placeholder checks (e.g. `**[Localized label: Programming language]:** [user choice]`) are well-scoped — consider pairing each English-leak check with enough surrounding template context so it only matches inside the template block, not anywhere in the file.

3. **`language.technical_terms` semantics** — the run-scoped state says "keep current semantics: `keep`". If the existing `config.yaml` has a non-`keep` value, the current wording is ambiguous about whether it is preserved or reset. Please clarify: preserve existing value if present, default to `keep` only when missing.

4. **Mode 2, Step 3 ordering reads backwards** — "Save to `.ai-factory/DESCRIPTION.md`. Write `.ai-factory/config.yaml` from … before saving this file." The instruction to write config appears after the save instruction it must precede. Reorder so the config-write instruction comes first, matching the actual execution order documented in Mode 1 Step 7.

5. **Final completion summary still contains English bullets** — in the "Architecture Generation" section the suggestions list (`/aif-docs — Generate project documentation`, etc.) is marked "Present … in resolved `language.ui`" but the bullet text itself stays in English. For consistency with the rest of the PR, mark those descriptions as `[Localized: …]` placeholders so the model does not anchor on the English strings.

6. **No behavioral test** — the new checks are static greps over `SKILL.md`. A regression where an artifact is written before `config.yaml`, or where the second setup run re-asks the language question, would not be caught. Out of scope for this PR, but worth tracking as a follow-up.

Happy to approve once 1 is addressed; the rest are polish.

ichinya:
What is closed:
- Explicitly fixed merge semantics for existing `config.yaml`: only resolved `language.ui` / `language.artifacts` are updated, and other sections are preserved.
- Clarified `language.technical_terms`: the existing value is preserved, and `keep` is written only if the key is missing.
- Reordered the Mode 2 wording so that the `config.yaml` entry comes before the `DESCRIPTION.md`.
- Removed the remaining English from existing-project suggestions and option labels, replacing them with localized placeholders.
- Made regression checks less fragile: the test now cuts out the required blocks of `SKILL.md` and checks English leaks and ordering within these sections, rather than using global `grep` across the entire file.

Verification:
- `npm test` passed completely: `55/55`.

lee-to (approved):

lee-to:
Fix conflict and merge

PR #82, opened 2026-04-19
fix: make extension install work on Windows and reject malformed add args

ichinya:
## Summary

Fix `ai-factory extension add` on Windows and make malformed input fail fast.

## Changes

- resolve npm execution in a Windows-safe way for npm-based extension installs
- add clearer missing-tool errors for `npm`, `git`, and `tar`
- reject excess arguments for `ai-factory extension add` instead of silently treating them as part of the source

## Why

Windows installs could fail with `spawnSync npm ENOENT` / `EINVAL`.
Also, duplicated input like:

`ai-factory extension add ai-factory extension add <url>`

was being interpreted incorrectly instead of returning a clear CLI error.

## Verification

- built the CLI with `npm run build`
- verified malformed `extension add` now fails with `too many arguments`
- verified npm-source path no longer crashes on Windows process spawning
- verified git-based extension install succeeds with a fixture extension repo

### DIFF
--- a/scripts/test-extensions.sh
+++ b/scripts/test-extensions.sh
@@ -1,0 +1,110 @@
+#!/bin/bash
+# Focused tests for extension resolver branches.
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+ROOT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+ROOT_DIR="$ROOT_DIR" node --input-type=module <<'EOF'
+import assert from 'node:assert/strict';
+import path from 'node:path';
+import { pathToFileURL } from 'node:url';
+
+const { resolveNpmCommand } = await import(pathToFileURL(path.join(process.env.ROOT_DIR, 'dist/core/extensions.js')).href);
+
+function createPathExists(paths) {
+  const existingPaths = new Set(paths);
+  return async targetPath => existingPaths.has(targetPath);
+}
+
+const execPath = path.join('runtime', 'bin', 'node.exe');
+const firstCandidate = path.join(path.dirname(execPath), 'node_modules', 'npm', 'bin', 'npm-cli.js');
+const secondCandidate = path.resolve(path.dirname(execPath), '..', 'node_modules', 'npm', 'bin', 'npm-cli.js');
+const thirdCandidate = path.resolve(path.dirname(execPath), '..', 'lib', 'node_modules', 'npm', 'bin', 'npm-cli.js');
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: '',
+    pathExists: createPathExists([firstCandidate]),
+  }),
+  { command: execPath, argsPrefix: [firstCandidate] },
+  'resolveNpmCommand must prefer npm-cli.js adjacent to the current node binary',
+);
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: '',
+    pathExists: createPathExists([secondCandidate]),
+  }),
+  { command: execPath, argsPrefix: [secondCandidate] },
+  'resolveNpmCommand must fall back to ../node_modules/npm/bin/npm-cli.js',
+);
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: '',
+    pathExists: createPathExists([thirdCandidate]),
+  }),
+  { command: execPath, argsPrefix: [thirdCandidate] },
+  'resolveNpmCommand must fall back to ../lib/node_modules/npm/bin/npm-cli.js',
+);
+
+const npmRoot = path.join('portable-npm');
+const npmCommandPath = path.join(npmRoot, 'npm.cmd');
+const npmCliPath = path.join(npmRoot, 'node_modules', 'npm', 'bin', 'npm-cli.js');
+const bundledNodePath = path.join(npmRoot, 'node.exe');
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: `ignored:${npmRoot}`,
+    pathDelimiter: ':',
+    pathExists: createPathExists([npmCommandPath, npmCliPath, bundledNodePath]),
+  }),
+  { command: bundledNodePath, argsPrefix: [npmCliPath] },
+  'resolveNpmCommand must resolve npm-cli.js from PATH on Windows when adjacent candidates are missing',
+);
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: npmRoot,
+    pathExists: createPathExists([npmCommandPath, npmCliPath]),
+  }),
+  { command: execPath, argsPrefix: [npmCliPath] },
+  'resolveNpmCommand must fall back to execPath when npm.cmd is found without bundled node.exe',
+);
+
+await assert.rejects(
+  () => resolveNpmCommand({
+    platform: 'win32',
+    execPath,
+    pathEnv: npmRoot,
+    pathExists: createPathExists([]),
+  }),
+  /safe Windows npm/i,
+  'resolveNpmCommand must fail closed when no safe Windows npm entrypoint exists',
+);
+
+assert.deepEqual(
+  await resolveNpmCommand({
+    platform: 'linux',
+    execPath,
+    pathEnv: '',
+    pathExists: createPathExists([]),
+  }),
+  { command: 'npm', argsPrefix: [] },
+  'resolveNpmCommand must fall back to npm on non-Windows platforms',
+);
+
+console.log('extension resolver unit tests passed');
+EOF
--- a/scripts/test-init.sh
+++ b/scripts/test-init.sh
@@ -358,6 +358,24 @@ const resolved = await resolveNpmCommand({
 assert.equal(resolved.command, bundledNodePath, 'Windows npm resolution must prefer node.exe adjacent to npm.cmd');
 assert.deepEqual(resolved.argsPrefix, [npmCliPath], 'Windows npm resolution must invoke npm-cli.js directly');
 
+const resolvedWithCustomDelimiter = await resolveNpmCommand({
+  platform: 'win32',
+  execPath: fakeExecPath,
+  pathEnv: `${path.relative(process.cwd(), npmRoot)}:${path.relative(process.cwd(), tempRoot)}`,
+  pathDelimiter: ':',
+});
+
+assert.equal(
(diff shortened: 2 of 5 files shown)


lee-to:
## Review

Overall the intent is good and the implementation is mostly clean. A couple of things worth addressing before merge:

### 🔴 Should fix

**1. `getPlatformCommand` is dead code** (`src/core/extensions.ts:115-117`)
```ts
function getPlatformCommand(command: 'git' | 'tar'): string {
  return command;
}
```
It returns its argument unchanged. Either drop it and inline `'git'` / `'tar'`, or make it do real work (e.g. mirror `resolveNpmCommand` and try to locate `git.exe` / `tar.exe` via `where` on Windows). As-is it's an abstraction that adds noise without value.

**2. cmd.exe fallback — potential argument injection** (`extensions.ts:134-139`)
```ts
return { command: process.env.ComSpec || 'cmd.exe', argsPrefix: ['/d', '/s', '/c', 'npm'] };
```
`execFileSync` with `cmd.exe /c` on Windows does not escape the subsequent arguments — cmd.exe parses them under its own rules (cf. CVE-2024-27980). If `packageName` ever contains `&`, `|`, `\"` or `%VAR%`, that becomes arbitrary command execution. Today the source is a CLI arg, so the user only hurts themselves, but for any future programmatic/registry flow the risk is real.
- Minimum: add a validation pass on `packageName` (whitelist `[a-zA-Z0-9@/._-]`) and/or leave a comment documenting the hazard.
- Better: in the Windows branch use `{ shell: true }` with `spawnSync`/`execFileSync` so Node applies its proper escaping.
- Best: when an `npm-cli.js` candidate is found, always prefer that path; keep the cmd.exe branch as a true last resort after an honest `where npm` lookup.

### 🟡 Minor

**3. No memoization for `resolveNpmCommand`** — three `fs.pathExists` calls per `extension add` is not a drama, but an easy win: cache the result at module scope.

**4. No unit tests for the new helpers** — `isMissingCommandError` / `isMissingNpmShellError` are trivial to cover with a small test that mocks `error.code` / `error.stderr`. The PR lists manual verification only.

**5. \"npm is required ... not found in PATH\" can mislead** — with the new resolver the failure may also come from a found `npm-cli.js` crashing due to a node/npm version mismatch. Consider pointing at `LOG_LEVEL=debug` in the message so users know where to look.

### 🟢 Nice

- `allowExcessArguments(false)` is a minimal, correct fix for the duplicated `extension add` issue.
- `tmpDir` cleanup on failure is now consistent across both resolvers — a real improvement over the previous behavior.
- The `npm-cli.js` candidate list covers the standard Node layouts (including the Windows `<node>\\node_modules\\npm\\bin\\` location).
- Error messages are clearer — users now learn which specific tool is missing.

### Verdict
Leaning **request changes** on the two items above: drop (or implement) `getPlatformCommand`, and close the cmd.exe escaping / `packageName` validation gap. The rest is nice-to-have.

ichinya:
Addressed the review points in this PR.

Changes:
- Removed the no-op `getPlatformCommand()` helper and inlined direct `git` / `tar` invocation.
- Reworked Windows npm resolution to avoid `cmd.exe` / shell fallback entirely.
  - We now resolve `npm-cli.js` directly.
  - First we try the usual locations near `process.execPath`.
  - If that is missing on Windows, we inspect `PATH` entries for `npm.cmd`, derive the matching `node_modules/npm/bin/npm-cli.js`, and run that directly.
  - If no safe `npm-cli.js` can be found, we fail closed with an explicit error instead of passing package specs through shell parsing.
- Added smoke coverage for:
  - `extension add` rejecting excess arguments
  - safe Windows npm resolution
  - explicit failure when no safe Windows npm entrypoint exists

This keeps full npm package-spec support intact while removing the shell-invocation risk from the Windows fallback path.

Verified locally with:
- `npm run build`
- Windows resolver smoke test against `dist/core/extensions.js`
- `node dist/cli/index.js extension add foo bar`
- local-path extension install smoke

lee-to:
## Follow-up review

Основной security-фикс (убран `cmd.exe`-фоллбэк, `npm-cli.js` теперь резолвится напрямую, fail-closed вместо шелла) сделан корректно. `getPlatformCommand` удалён, cleanup `tmpDir` теперь симметричен для npm/git/tar, `resolveNpmCommand` стал инжектируемым — хорошая заявка на тесты.

Но PR мержить нельзя — **CI красный из-за нового smoke-теста**, и это симптом реального бага в резолвере.

### 🔴 Blocker — cross-platform delimiter bug в `resolveNpmCommand`

`src/core/extensions.ts:121` — `findWindowsPathNpmCli` делает:
```ts
pathEnv.split(path.delimiter)
```
`path.delimiter` берётся с **хост-системы**, а не из переданного `options.platform`. На Ubuntu CI это `:`, а тест в `scripts/test-init.sh` формирует PATH через `;` (Windows-стиль):
```
pathEnv: `${npmRoot};${process.env.PATH}`
```
Итог из свежего CI-прогона ([run 24671086859](https://github.com/lee-to/ai-factory/actions/runs/24671086859/job/72141771190)):
```
Error: Unable to locate npm-cli.js for a safe Windows npm invocation.
    at resolveNpmCommand (dist/core/extensions.js:84:15)
```
Позитивный сценарий не находит `npm.cmd` → ветка `findWindowsPathNpmCli` возвращает `null` → бросается «fail-closed» ошибка. Негативный ассерт проходит случайно (по той же причине — ничего не находится).

В проде баг менее заметен (Windows runtime = `path.delimiter === ';'`), но он реальный: функция принимает `platform` как аргумент, а фактически игнорирует его для разбора `PATH`/путей. Это и делает тест не воспроизводимым на Linux/macOS.

**Фикс** — один из:
- `const delimiter = platform === 'win32' ? ';' : ':'; pathEnv.split(delimiter)` (минимальная правка).
- Пробросить `delimiter` в `ResolveNpmCommandOptions` и использовать в `findWindowsPathNpmCli`.

Без этого новый тест — фиктивный: на Windows он проходит из-за совпадения делимитера, на unix CI падает. Как только поправите — обе проверки (позитивная и «safe Windows npm» отказ) начнут тестировать ровно то, что заявлено.

### 🟡 Minor

1. `fs.pathExists` по 3 кандидатам + walk по `PATH` дёргается на **каждый** `extension add`. Результат детерминирован в рамках процесса — можно мемоизировать на уровне модуля (один раз на жизнь CLI).
2. Сообщение `'npm is required … not found in PATH'` теперь может сработать и когда `npm-cli.js` найден, но упал по другой причине (version mismatch и т.п.). Стоит упомянуть `LOG_LEVEL=debug` для диагностики.
3. Smoke-тесты неплохи как интеграция, но `isMissingCommandError` и ветки `resolveNpmCommand` (все три кандидата, win32-PATH-walk, fail-closed) просятся в юнит-тесты с моком `fs.pathExists` — быстрее и покрывают платформенные пути без трюков с `platform: 'win32'` на Linux.

### Verdict
**Request changes** по пункту 1 — CI должен быть зелёным, и исправление мелкое. После фикса делимитера PR можно мержить.

ichinya:
Addressed the blocker from the follow-up review.

Changes:
- `resolveNpmCommand()` no longer relies on the host `path.delimiter` when resolving Windows npm paths.
- Added explicit `pathDelimiter` handling and derive the default delimiter from the target `platform` (`win32 -> ';'`, otherwise `':'`).
- Updated the Windows smoke coverage to include an injected custom-delimiter case, so the test now exercises the cross-platform path parsing instead of accidentally depending on host defaults.

Local verification:
- `npm run build`
- `npm run lint`
- direct `resolveNpmCommand` smoke for:
  - normal Windows-style `;` PATH
  - injected custom delimiter
  - fail-closed path when no safe `npm-cli.js` is available
- `node dist/cli/index.js extension add foo bar`
- local-path extension install smoke

This should address the Ubuntu CI failure around the Windows npm resolver path splitting.

lee-to (approved):
## Follow-up review

Blocker from the previous round is resolved — CI is green on `461aac9`.

### ✅ Fixed
- `resolveNpmCommand` no longer leaks `path.delimiter` from the host. `getPathDelimiter(platform, override)` derives `;` / `:` from the target `platform`, and `findWindowsPathNpmCli` takes the delimiter explicitly.
- `ResolveNpmCommandOptions.pathDelimiter` is injectable, and the smoke test now covers an injected custom-delimiter case — so the Windows PATH-walk is actually exercised on Ubuntu CI instead of accidentally passing.
- Symmetric `tmpDir` cleanup across npm/git/tar branches is preserved, shell fallback stays removed, fail-closed path is intact.

### 🟡 Minor (non-blocking, fine as follow-up)
1. `resolveNpmCommand` still runs `fs.pathExists` on every `extension add`. Result is deterministic per process — worth memoizing at module scope.
2. `'npm is required ... not found in PATH'` can also fire when `npm-cli.js` is found but crashes for other reasons (version mismatch, etc.). Mentioning `LOG_LEVEL=debug` would help users self-diagnose.
3. Smoke coverage is solid, but `resolveNpmCommand` branches (all three candidates, win32 PATH-walk, fail-closed) are still easier to pin down with unit tests mocking `fs.pathExists`.

### Verdict
Approving — the security fix is correct, the cross-platform regression is closed, and the remaining items are nice-to-haves.

ichinya:
Addressed the remaining follow-up items from the approved review.

Changes:
- memoized `resolveNpmCommand()` for stable process-wide resolution instead of re-running the same filesystem checks on every `extension add`
- improved npm failure diagnostics with `LOG_LEVEL=debug` hints, including the case where a safe `npm-cli.js` entrypoint is resolved but the command still cannot start
- added focused resolver coverage for:
  - all three adjacent `npm-cli.js` candidates
  - Windows PATH-walk resolution
  - fallback to `execPath` when `npm.cmd` exists without bundled `node.exe`
  - fail-closed Windows behavior
  - non-Windows fallback to plain `npm`
- wired the new focused resolver test into `npm test`

Verified locally with:
- `npm run build`
- `npm run lint`
- focused resolver checks against `dist/core/extensions.js`
- `node dist/cli/index.js extension add foo bar`
- local-path extension install smoke

This is follow-up hardening only; no change to the approved security direction (`cmd.exe` fallback remains removed, direct `npm-cli.js` execution remains intact).

lee-to:
Follow-up items all look good on `28a2ec7`, CI green.

- **Memoization:** `npmCommandResolutionCache` keys on `[platform, execPath, pathEnv, pathDelimiter]`, and explicitly bypasses the cache when a custom `pathExists` is injected — so unit tests don't poison the shared Map. Clean.
- **`LOG_LEVEL=debug` hints:** split into two branches — one for the "safe entrypoint resolved but couldn't start" case (version mismatch, etc.), one for the plain "npm not in PATH" case. That's exactly the ambiguity I was worried about.
- **Focused resolver tests:** `scripts/test-extensions.sh` covers all three adjacent candidates, Windows PATH-walk, the `npm.cmd` without bundled `node.exe` fallback, fail-closed, and the non-Windows `npm` fallback. Wired into `npm test` via `test-skills.sh`.

Nothing else from my side — good to merge whenever.

PR #101, opened 2026-04-28
feat(aif-plan): honor HANDOFF_BRANCH_PREPARED contract

### LINKED ISSUE #96: aif-plan: HANDOFF_MODE=1 + full mode skips Step 1.4 branch creation
Related downstream bug: https://github.com/lee-to/aif-handoff/issues/83
Downstream runtime fix: https://github.com/lee-to/aif-handoff/pull/85

## Problem

When `/aif-plan` runs under `HANDOFF_MODE=1` (autonomous Handoff agent) with `mode: full` and `git.create_branches: true`, the skill does **not** execute Step 1.4 (branch/worktree creation). Instead, it writes only a suggested branch name into the plan text. Result: autonomous tasks keep writing into whatever branch happens to be checked out, so multiple queued tasks pile their diffs into the same worktree.

## Suspected cause

Section `### When HANDOFF_MODE is 1` in `aif-plan/SKILL.md` states:

> - **No interactive questions:** Do not use AskUserQuestion — use sensible defaults (verbose logging, yes to tests, yes to docs, skip roadmap linkage).

The LLM appears to read this as "skip everything that would otherwise need user input" and bails out before reaching Step 1.4 (which does not actually need user input — it only reads config values). Same behavior is visible in `.claude/agents/plan-polisher.md` branch-creation rules, which are similarly conditional and easy to skip.

## New contract (after aif-handoff#85)

Handoff now owns branch creation at the agent-code level. The skill must no longer execute `git checkout -b` when running under Handoff — otherwise we get double-branching (Handoff creates `feature/foo-abc123`, skill creates `feature/foo`, plan file gets written on the skill's branch, Handoff's post-run assertion fires `branch_drift` and the task is pinned to `blocked_external`).

The Handoff planner injects these prompt lines:

```
HANDOFF_MODE: 1
HANDOFF_TASK_ID: <uuid>
HANDOFF_BRANCH_PREPARED: 1
HANDOFF_BRANCH_NAME: feature/<slug>-<short-id>
```

## Requested fix (skill-only)

**`aif-plan/SKILL.md`** — replace the HANDOFF_MODE=1 section to add ownership rules:

```
If HANDOFF_BRANCH_PREPARED=1:
  - Do NOT execute git checkout, git pull, or git checkout -b.
  - Do NOT create a worktree.
  - Read HANDOFF_BRANCH_NAME from the prompt.
  - Validate that `git rev-parse --abbrev-ref HEAD` === HANDOFF_BRANCH_NAME.
  - If it does not match, STOP and report a blocker in the plan summary:
      "Branch drift: expected <HANDOFF_BRANCH_NAME>, actual <current>."
    Do NOT "fix" drift by switching or creating.
  - Use HANDOFF_BRANCH_NAME as the plan filename stem
    (e.g. `.ai-factory/plans/<HANDOFF_BRANCH_NAME-with-slashes-replaced>.md`).

If HANDOFF_MODE=1 but HANDOFF_BRANCH_PREPARED is unset or 0:
  - Fallback: execute Step 1.4 branch creation normally.
  - This path handles older Handoff clients that haven't adopted
    the prepared-branch contract yet.

If HANDOFF_MODE is unset (manual/skill-only runs):
  - Current behavior: Step 1.4 creates/switches branches per config.
```

**`plan-polisher.md`** — replace the `Branch creation (full mode only)` block with the same `HANDOFF_BRANCH_PREPARED` gate. Also remove the heuristic ""if current branch contains `/`, use it as-is"" — under `HANDOFF_BRANCH_PREPARED=1` the skill must compare for strict equality against `HANDOFF_BRANCH_NAME`, and false positives on unrelated branches like `release/v1` must not be accepted silently.

## Why the earlier wording in this issue was dangerous

The original suggestion "you MUST execute the \`git checkout -b <branch>\` block when git.create_branches=true" would collide with Handoff's prepared branch: skill would create a second branch, abandoning Handoff's plan file mid-flight. The contract above keeps the skill correct both standalone AND under Handoff.

## Downstream enforcement

Handoff (aif-handoff#85 + follow-up) now also:
- validates `branchName` via `git check-ref-format --branch`
- treats `task.branchName` as source-of-truth (`restorePersistedBranch` throws rather than silent-skip when config drifts off between stages)
- `assertCurrentBranch` runs post-subagent in planner, implementer, plan-checker, reviewer
- classifies `BranchIsolationError` as `blocked_external` with `retryAfter=null` (no silent retry)

So even without this skill-level fix, Handoff stops instead of corrupting state. But users running the skill standalone (without Handoff) still benefit from the unconditional Step 1.4 clarification.

## Related

- aif-handoff#83 — original bug
- aif-handoff#85 — agent-code fix
- aif-handoff#86 — per-task worktrees (parallel isolation follow-up)

lee-to:
## Summary
- `aif-plan` and `plan-polisher` now respect Handoff's prepared-branch contract: when `HANDOFF_BRANCH_PREPARED=1`, the skill no longer runs `git checkout -b`, validates strict equality against `HANDOFF_BRANCH_NAME`, and uses that name as the plan filename stem.
- On branch drift, the skill STOPS and reports a blocker instead of "fixing" by switching/creating — matching the `BranchIsolationError` / `blocked_external` classification added in aif-handoff#85.
- Fallback path preserved for older Handoff clients (HANDOFF_MODE=1 without `HANDOFF_BRANCH_PREPARED`) and standalone `/aif-plan` runs (Step 1.4 still owns branch creation per `git.create_branches`).

Closes #96. Pairs with [aif-handoff#85](https://github.com/lee-to/aif-handoff/pull/85).

## Test plan
- [ ] Standalone `/aif-plan full <desc>` on `main` with `git.create_branches: true` still creates `feature/<slug>` and writes `<plans>/feature-<slug>.md`
- [ ] Standalone `/aif-plan full <desc>` with `git.create_branches: false` skips branch ops and writes `<plans>/<slug>.md`
- [ ] Handoff run injects `HANDOFF_MODE=1 HANDOFF_BRANCH_PREPARED=1 HANDOFF_BRANCH_NAME=feature/foo-abc123` → skill writes `<plans>/feature-foo-abc123.md` on the prepared branch, no `git checkout -b`, no double-branch
- [ ] Drift case (Handoff prepares `feature/foo-abc123`, current branch is `main`) → skill reports `Branch drift: expected ..., actual ...` blocker and does not switch
- [ ] Legacy Handoff client (`HANDOFF_MODE=1`, no `HANDOFF_BRANCH_PREPARED`) → fallback to Step 1.4 branch creation per config
- [ ] `plan-polisher` subagent receives same env/prompt vars and applies identical gating

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/skills/aif-plan/SKILL.md
+++ b/skills/aif-plan/SKILL.md
@@ -330,7 +330,7 @@ Docs policy semantics:
 
 - Skip this entire step. Branch validation already happened in Step 0.
 - The plan file path uses `HANDOFF_BRANCH_NAME` (slashes replaced by `-`) as the stem.
-- Do NOT run `git checkout`, `git pull`, `git checkout -b`, or `git worktree add`.
+- Do **NOT** run `git checkout`, `git pull`, `git checkout -b`, or `git worktree add`.
 
 **If `git.enabled = false` or `git.create_branches = false`:**
 
--- a/skills/aif-plan/SKILL.md
+++ b/skills/aif-plan/SKILL.md
@@ -18,7 +18,14 @@ Create an implementation plan for a feature or task. Two modes:
 
 ### Step 0 (pre): Detect Handoff Mode
 
-Determine Handoff mode, task ID, and branch contract. If the caller passed `HANDOFF_MODE`, `HANDOFF_TASK_ID`, `HANDOFF_BRANCH_PREPARED`, and `HANDOFF_BRANCH_NAME` as explicit text in the prompt, use those values. Otherwise, use the Bash tool to read the environment variables:
+Determine Handoff mode, task ID, and branch contract. Resolve each value independently so legacy callers that pass only `HANDOFF_MODE` and `HANDOFF_TASK_ID` still enter Handoff mode correctly:
+
+- `HANDOFF_MODE`: explicit prompt value if present; otherwise environment value; otherwise empty string.
+- `HANDOFF_TASK_ID`: explicit prompt value if present; otherwise environment value; otherwise empty string.
+- `HANDOFF_BRANCH_PREPARED`: explicit prompt value if present; otherwise environment value; otherwise `0`.
+- `HANDOFF_BRANCH_NAME`: explicit prompt value if present; otherwise environment value; otherwise empty string.
+
+Use the Bash tool only for values that were not passed explicitly in the prompt:
 
 ```
 Bash: printenv HANDOFF_MODE || true
@@ -44,7 +51,8 @@ Handoff owns branch creation at the agent-code level. The skill must NOT create 
 **If `HANDOFF_BRANCH_PREPARED` is `1`:**
 
 - Do **NOT** execute `git checkout`, `git pull`, or `git checkout -b`.
-- Do **NOT** create a worktree (ignore `--parallel` for branch creation purposes).
+- Treat `--parallel` as disabled for all downstream behavior.
+- Do **NOT** create a worktree.
 - Read `HANDOFF_BRANCH_NAME` from the prompt / env.
 - Validate strict equality:
   ```
@@ -331,6 +339,7 @@ Docs policy semantics:
 - Skip this entire step. Branch validation already happened in Step 0.
 - The plan file path uses `HANDOFF_BRANCH_NAME` (slashes replaced by `-`) as the stem.
 - Do **NOT** run `git checkout`, `git pull`, `git checkout -b`, or `git worktree add`.
+- Treat `--parallel` as disabled: do not create a worktree and do not auto-invoke `/aif-implement`.
 
 **If `git.enabled = false` or `git.create_branches = false`:**
 
@@ -542,7 +551,7 @@ Use the canonical template in `references/TASK-FORMAT.md` (Plan File Template).
 
 ### Step 6: Next Steps
 
-**Full mode + parallel (`--parallel`):** Automatically invoke `/aif-implement` — the whole point of parallel is autonomous end-to-end execution in an isolated worktree.
+**Full mode + parallel (`--parallel`):** Automatically invoke `/aif-implement` — the whole point of parallel is autonomous end-to-end execution in an isolated worktree. If `HANDOFF_BRANCH_PREPARED = 1`, treat `--parallel` as disabled and do not auto-invoke `/aif-implement`.
 
 ```
 /aif-implement
--- a/skills/aif-plan/SKILL.md
+++ b/skills/aif-plan/SKILL.md
@@ -18,11 +18,13 @@ Create an implementation plan for a feature or task. Two modes:
 
 ### Step 0 (pre): Detect Handoff Mode
 
-Determine Handoff mode and task ID. If the caller passed `HANDOFF_MODE` and `HANDOFF_TASK_ID` as explicit text in the prompt, use those values. Otherwise, use the Bash tool to read the environment variables:
+Determine Handoff mode, task ID, and branch contract. If the caller passed `HANDOFF_MODE`, `HANDOFF_TASK_ID`, `HANDOFF_BRANCH_PREPARED`, and `HANDOFF_BRANCH_NAME` as explicit text in the prompt, use those values. Otherwise, use the Bash tool to read the environment variables:
 
 ```
 Bash: printenv HANDOFF_MODE || true
 Bash: printenv HANDOFF_TASK_ID || true
+Bash: printenv HANDOFF_BRANCH_PREPARED || true
+Bash: printenv HANDOFF_BRANCH_NAME || true
 ```
 
 **Then check `HANDOFF_MODE`:**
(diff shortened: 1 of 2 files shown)


ichinya:
**Verdict: Request changes — one correctness blocker, otherwise the contract looks aligned.**

The main prepared-branch behavior is implemented in the right direction: when `HANDOFF_BRANCH_PREPARED=1`, the skill now forbids branch/worktree creation, validates `git rev-parse --abbrev-ref HEAD` against `HANDOFF_BRANCH_NAME`, reports drift as a blocker, and derives the plan filename from the prepared branch name. That matches the requested contract in #96, including the requirement to preserve a fallback path for older Handoff clients. 

### Blocker: explicit Handoff variable parsing became all-or-nothing

In `skills/aif-plan/SKILL.md`, the new Step 0 wording says to use explicit prompt values only if the caller passed **all four** values: `HANDOFF_MODE`, `HANDOFF_TASK_ID`, `HANDOFF_BRANCH_PREPARED`, and `HANDOFF_BRANCH_NAME`; otherwise it falls back to reading all values from the environment. 

That can break the legacy path this PR explicitly tries to preserve. #96 says older clients may run with `HANDOFF_MODE=1` while `HANDOFF_BRANCH_PREPARED` is unset or `0`, and in that case the skill should execute Step 1.4 normally. But if an older caller passes only the legacy explicit prompt fields (`HANDOFF_MODE`, `HANDOFF_TASK_ID`) and does not set env vars, the current all-or-nothing wording can cause the skill to ignore the explicit `HANDOFF_MODE=1` and fall into manual-mode behavior instead of the intended legacy fallback.

Same concern applies to `subagents/plan-polisher.md`: it now documents all four prompt variables and then says to fall back to env “if the caller did NOT pass them,” which is ambiguous in the same all-or-nothing way.

Suggested fix: parse each variable independently.

```text
HANDOFF_MODE =
  explicit HANDOFF_MODE if present
  else env HANDOFF_MODE
  else ""

HANDOFF_TASK_ID =
  explicit HANDOFF_TASK_ID if present
  else env HANDOFF_TASK_ID
  else ""

HANDOFF_BRANCH_PREPARED =
  explicit HANDOFF_BRANCH_PREPARED if present
  else env HANDOFF_BRANCH_PREPARED
  else "0"

HANDOFF_BRANCH_NAME =
  explicit HANDOFF_BRANCH_NAME if present
  else env HANDOFF_BRANCH_NAME
  else ""
```

Then branch as currently intended:

```text
if HANDOFF_MODE == "1" and HANDOFF_BRANCH_PREPARED == "1":
  validate strict branch equality, no checkout/pull/worktree

elif HANDOFF_MODE == "1":
  legacy fallback: Step 1.4 branch creation per git.create_branches

else:
  manual behavior
```

### Non-blocking clarification: `--parallel` should probably be disabled downstream too

The PR says to ignore `--parallel` for branch creation and skip worktree creation under `HANDOFF_BRANCH_PREPARED=1`. However, `aif-plan` still has downstream full-mode + parallel behavior that automatically invokes `/aif-implement` after planning. If `--parallel` ever reaches this path under Handoff-prepared mode, the worktree is skipped but the auto-implement behavior could still run in the prepared branch/root worktree.

I’d clarify that `HANDOFF_BRANCH_PREPARED=1` treats `--parallel` as disabled for **all** downstream behavior, not just branch/worktree creation.

### Minor: hidden/bidirectional Unicode warning

GitHub shows hidden/bidirectional Unicode warnings for both changed markdown files. Since these are instruction files and the second commit is specifically working around scanner behavior, it would be safer to either remove the hidden characters or confirm they are pre-existing/intentional.

### Positive notes

The core contract is otherwise well covered: strict equality against `HANDOFF_BRANCH_NAME`, no partial/prefix/“contains slash” heuristics, no drift self-healing, and prepared-branch filename derivation are all present in both `aif-plan` and `plan-polisher`.  The single CI job also succeeded.

lee-to:
Addressed the review feedback in 8765ea9.

Changes made:
- Updated `skills/aif-plan/SKILL.md` to resolve each Handoff variable independently: explicit prompt value first, then environment, then the documented default. This preserves the legacy Handoff path where older callers pass only `HANDOFF_MODE` and `HANDOFF_TASK_ID`.
- Mirrored the same independent variable-resolution contract in `subagents/plan-polisher.md`.
- Clarified that `HANDOFF_BRANCH_PREPARED=1` treats `--parallel` as disabled for all downstream behavior, including no worktree creation and no automatic `/aif-implement` invocation.
- Checked for hidden/bidirectional control characters in the touched markdown files; none were found.

Validation:
- `git diff --check` passed.
- `npm test` validated all skills successfully, including `aif-plan`, then failed later in the existing gate-result smoke script with `scripts/test-gate-result-contract.sh: line 183: verify: unbound variable`. That failure appears unrelated to this PR change set; it occurs in the test script's associative-array section after the skill validation, update/init smoke tests, aif-qa smoke tests, and aif-rules-check smoke tests had passed.

ichinya:
Verdict: Approve

Re-reviewed after `8765ea9`. The previous blocker around all-or-nothing Handoff variable parsing is fixed: `aif-plan` now resolves `HANDOFF_MODE`, `HANDOFF_TASK_ID`, `HANDOFF_BRANCH_PREPARED`, and `HANDOFF_BRANCH_NAME` independently, using explicit prompt value first, then env, then the documented default. `plan-polisher` mirrors the same contract, so legacy Handoff callers that only pass `HANDOFF_MODE` / `HANDOFF_TASK_ID` still enter the fallback path correctly.

The prepared-branch contract now matches #96:

- `HANDOFF_BRANCH_PREPARED=1` forbids `git checkout`, `git pull`, `git checkout -b`, and worktree creation.
- Current branch is validated by strict equality against `HANDOFF_BRANCH_NAME`.
- Branch drift is reported as a blocker instead of being “fixed” by switching or creating branches.
- Plan filename stem comes from `HANDOFF_BRANCH_NAME` with slashes replaced.
- `HANDOFF_MODE=1` with `HANDOFF_BRANCH_PREPARED` unset or `0` still falls back to normal Step 1.4 branch creation for older clients.

The earlier `--parallel` concern is also addressed: prepared-branch mode now treats `--parallel` as disabled for all downstream behavior, including no worktree creation and no automatic `/aif-implement` invocation.

CI is green on the latest head. The remaining GitHub Actions Node.js 20 deprecation warning is unrelated to this change set.

LGTM.

PR #102, opened 2026-04-29
feat(schema): publish extension manifest JSON schema

### LINKED ISSUE #86: feat(schema): publish extension.json schema
re-scope of existing draft #34.

PR #34 already tries to add JSON schemas for `.ai-factory.json` and `extension.json`, but it is in draft and has received comments: draft 2020-12 needs `$defs`, schemas should be included in package, it is difficult to validate strictly `.ai-factory.json` as internal runtime state, and `extension.json` looks like the best schema target.

**Why:** after #71 `extension.json` has become more important: it contains runtime definitions, `agentFiles`, `replaces`, `injections`, `commands`, and MCP templates. For `aifhub-extension`, this will provide early validation of the manifest before install.

 **Suggested scope:**

* Narrow the PR to only `extension.schema.json`.
* Add a `$schema` reference in the example extension manifests.
* Add an `agentFiles` schema:

  * `runtime: string`;
  * `source: string`;
  * `target: string`;
  * required all three.
* Add `agents[].agentsDir' + `agentFileExtension'.
* Add the package publish config so that schema actually gets into the npm tarball.
* Do not validate `.ai-factory.json` in this PR.

**Files:**

```text
schemas/extension.schema.json
examples/extensions/aif-ext-hello/extension.json
docs/extensions.md
package.json
scripts/test-*.sh, if there is a package smoke
```

**Acceptance criteria:**

* Schema matches the current `ExtensionManifest`.
* `agentFiles` are described by runtime-aware objects.
* `$defs`, not `definitions`, if using JSON Schema 2020-12.
* Schema is published in the npm package.
* The example extension passes validation.
* `.ai-factory.json`json` is not touched in this PR.

**Tests:**

* `npm pack --dry-run` includes schema;
* JSON schema validation example;
* negative example: `agentFiles` as an array of strings fails.

ichinya:
## Summary

Closes #86.

- Added `schemas/extension.schema.json` for public `extension.json` manifests using JSON Schema draft 2020-12 and `$defs`.
- Covered runtime-aware fields including `agentFiles`, `agents[].agentsDir`, `agents[].agentFileExtension`, and MCP server templates.
- Added `$schema` to the example extension manifest.
- Added Ajv-based schema smoke tests for the positive example and negative `agentFiles: ["bad"]` case.
- Ensured the schema is included in the npm package via `package.json.files` and `npm pack --dry-run --json`.
- Documented schema usage and the boundary between schema validation and AI Factory runtime semantic checks.

## Scope Notes

This intentionally targets `extension.json` only. It does not add or edit `.ai-factory.json` schema/config validation.

## Verification

- `npm run build`
- `npm run test:extensions`
- `npm run lint`
- `npm pack --dry-run --json`
- `git diff --check`
- `npm audit --omit=dev`

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -166,7 +166,7 @@ my-extension/
 
 ```json
 {
-  "$schema": "./node_modules/ai-factory/schemas/extension.schema.json",
+  "$schema": "https://raw.githubusercontent.com/lee-to/ai-factory/2.x/schemas/extension.schema.json",
   "name": "aif-ext-example",
   "version": "1.0.0",
   "description": "Example extension",
@@ -228,7 +228,23 @@ my-extension/
 
 Only `name` and `version` are required. All other fields are optional.
 
-AI Factory publishes the extension manifest schema at `schemas/extension.schema.json` in the npm package. Extension authors may add an optional `$schema` property for editor/tooling validation; use a relative path that works from your manifest location, or the schema URI when your tooling resolves it.
+AI Factory publishes the extension manifest schema at `schemas/extension.schema.json` in the npm package. Extension authors may add an optional `$schema` property for editor/tooling validation.
+
+Use the public schema URL when the extension should work regardless of whether the `ai-factory` CLI is installed locally or globally:
+
+```json
+{
+  "$schema": "https://raw.githubusercontent.com/lee-to/ai-factory/2.x/schemas/extension.schema.json"
+}
+```
+
+If your extension project installs `ai-factory` as a local dependency and you want offline or package-version-aligned validation, use the package-local schema path instead, adjusting the relative path from your `extension.json` location:
+
+```json
+{
+  "$schema": "./node_modules/ai-factory/schemas/extension.schema.json"
+}
+```
 
 The schema validates the public manifest shape: commands, runtime definitions, runtime-aware `agentFiles`, injections, skills, replacements, and MCP server templates. Install-time semantics still run in AI Factory itself, including runtime existence, safe relative paths, extension matching, duplicate ownership, and whether a runtime supports managed agent files.
 
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -166,6 +166,7 @@ my-extension/
 
 ```json
 {
+  "$schema": "./node_modules/ai-factory/schemas/extension.schema.json",
   "name": "aif-ext-example",
   "version": "1.0.0",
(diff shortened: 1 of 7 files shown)


lee-to:
Review note:

I think this needs one adjustment before merge for the global-install case.

`docs/extensions.md` now recommends:

```json
"$schema": "./node_modules/ai-factory/schemas/extension.schema.json"
```

That works only when `ai-factory` is installed locally in the project/extension. If the CLI is installed globally, there is no project-local `node_modules/ai-factory`, so editor/tooling schema resolution fails.

The docs also mention using the schema URI, but the schema `$id` is currently `https://ai-factory.dev/schemas/extension.schema.json`, and that host does not resolve from here (`curl -I https://ai-factory.dev/schemas/extension.schema.json` -> `Could not resolve host`). So for global installs there is currently no documented `$schema` value that reliably works.

Suggested fix: either host the public schema URI and document it as the recommended `$schema`, or document local/global cases separately and provide/copy a stable project-local schema path during setup/update.

I verified `npm_config_cache=/tmp/aif-npm-cache npm run test:extension-schema` passes; the default run only failed because this machine has a root-owned npm cache under `/Users/macos/.npm`.

ichinya:
Addressed.

I changed the schema `$id` to a public GitHub raw URL and made that URL the recommended `$schema` value in the docs, so it does not depend on a project-local `node_modules/ai-factory` install. The package-local `./node_modules/ai-factory/schemas/extension.schema.json` path is now documented only as an offline/version-aligned option for extension projects that install `ai-factory` locally.

I also added schema smoke coverage to keep the public `$id` and the local/global docs distinction from regressing.

lee-to:
Thanks, this addresses my review note. The docs now recommend a schema URL that does not depend on a project-local `node_modules`, and the local package path is framed as the offline/version-aligned option for projects that install `ai-factory` locally.

Verified on the PR branch:
- `npm_config_cache=/tmp/aif-npm-cache npm run test:extension-schema` passes
- `npm_config_cache=/tmp/aif-npm-cache npm run build` passes
- `npm_config_cache=/tmp/aif-npm-cache npm run test:extensions` passes after build

One expected pre-merge caveat: `https://raw.githubusercontent.com/lee-to/ai-factory/2.x/schemas/extension.schema.json` returns 404 right now because the schema is not on `2.x` until this PR is merged. That is not a blocker for the merged state.

PR #106, opened 2026-05-07
docs(aif-architecture): enhance DDD and vertical slice architecture guidance

DAGpro:
Add comprehensive Domain-Driven Design examples with bounded contexts, 
CQRS patterns, and vertical slice organization.
Include detailed folder structures, core principles, and anti-patterns for better architectural decision-making.

### DIFF
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -255,12 +255,13 @@ Typical sequence when both are useful:
 
 For the workflow view of where these fit, see [Development Workflow](workflow.md).
 
-### `/aif-architecture [clean|ddd|microservices|monolith|layers]`
+### `/aif-architecture [structured|explicit|vertical|microservices|layers]`
 Generates architecture guidelines tailored to your project:
 ```
 /aif-architecture           # Analyze project and recommend
-/aif-architecture clean     # Use Clean Architecture
-/aif-architecture monolith  # Use Modular Monolith
+/aif-architecture explicit  # Use Explicit Architecture
+/aif-architecture vertical  # Use Vertical Slice + Explicit Architecture
+/aif-architecture structured # Use Structured Modules
 ```
 - Reads the resolved description artifact for project context
 - Recommends architecture pattern based on team size, domain complexity, and tech stack
--- a/docs/skills.md
+++ b/docs/skills.md
@@ -255,14 +255,18 @@ Typical sequence when both are useful:
 
 For the workflow view of where these fit, see [Development Workflow](workflow.md).
 
-### `/aif-architecture [structured|explicit|vertical|microservices|layers]`
+### `/aif-architecture [explicit|structured|microservices|layers]`
 Generates architecture guidelines tailored to your project:
 ```
 /aif-architecture           # Analyze project and recommend
-/aif-architecture explicit  # Use Explicit Architecture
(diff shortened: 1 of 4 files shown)


lee-to:
Thanks for the PR. I reviewed both the skill behavior contract and the proposed DDD/architecture content.

I would not merge this as-is. The direction is useful, but a few parts need tightening first:

1. Please move the detailed DDD / Explicit Architecture material out of `skills/aif-architecture/SKILL.md` and into references, for example `skills/aif-architecture/references/explicit-architecture.md` or `ddd.md`. The skill itself should stay small: decision matrix, selection logic, and an instruction like "if the user chooses Explicit/DDD/Vertical Slice, read the matching reference before generating `ARCHITECTURE.md`". This keeps the prompt smaller and avoids turning the skill into a long embedded handbook.

2. The public argument contract changed from `clean|ddd|microservices|monolith|layers` to `structured|explicit|vertical|microservices|layers` in both the frontmatter and docs. That removes documented options (`clean`, `ddd`, `monolith`) rather than adding new guidance. If this is intentional, it needs a migration/backward-compatibility note and probably aliases. If not, keep the old options and add the new ones as additional supported aliases.

3. The DDD description is not quite accurate yet. In particular:
   - `Explicit Architecture` is described as a fusion of DDD, Hexagonal, Onion, Clean, and CQRS. Those are related but not interchangeable; CQRS is optional, and DDD is primarily about modeling and boundaries, not a folder structure.
   - The text says dependencies "point outward" while also saying inner layers never import outer layers. For Clean/Onion/Hexagonal style architecture, source dependencies should point inward toward the domain/application core; adapters/framework code sits outside and depends on the inside contracts.
   - "A bounded context should be extractable into its own microservice without internal code changes" is too strong. A bounded context can be a good service boundary candidate, but extraction usually requires operational, persistence, integration, deployment, and transaction-boundary work.
   - `ContextMap/` is labeled as an anti-corruption layer. A Context Map is a strategic DDD relationship map; an ACL is one possible integration pattern/adapter. They should not be presented as the same mandatory folder.
   - CQRS should not be mandatory for DDD/Explicit Architecture. It should be recommended only when read/write models or use-case complexity justify it.

4. The new `Structured Modules` path is a useful pragmatic option, but it currently claims business logic lives in services while models merely contain some validation/behavior. That is fine as a simpler modular style, but it should be clearly separated from DDD so users do not mistake service-centric modules for a domain model.

Validation result: I ran `npm test`. The suite reached the final gate contract smoke test and failed with `scripts/test-gate-result-contract.sh:183: verify: unbound variable`. This PR only changes `docs/skills.md` and `skills/aif-architecture/SKILL.md`, so that failure does not look caused by this diff, but the branch is not currently fully green in my local run.

Suggested shape for the next revision:
- Keep `/aif-architecture` compact.
- Add architecture references under `skills/aif-architecture/references/`.
- Preserve or alias existing options (`clean`, `ddd`, `monolith`) unless this PR intentionally introduces a breaking CLI/skill contract change.
- Rework the DDD reference around actual DDD concepts: ubiquitous language, bounded contexts, aggregates/invariants, value objects, domain events, repositories as domain-facing abstractions, context mapping patterns, and optional integration with hexagonal/clean architecture and CQRS.

PR #108, opened 2026-05-10
feat(aif-plan): add sequential numbered plan-file format

TimurTurdyev:
## Summary

Adds an opt-in `sequential` value for `workflow.plan_id_format` so that
`/aif-plan full` writes plan files with a 4-digit numeric prefix
(`<NNNN>_<stem>.md`) under `paths.plans/`. Mirrors the plan-file convention
already used in real-world projects (e.g. mercurioplatform).

The default `slug` behavior is unchanged. `timestamp` and `uuid` remain
reserved values and currently fall back to `slug` with an `INFO` log —
they are not active formats in this PR.

## How it works

- `stem` is the **canonical** `plan_file_stem` chosen by `/aif-plan`
  Step 1.2.a — Handoff branch (slashes → `-`) → generated git branch
  (slashes → `-`) → description slug. Branch-based consumers
  (`/aif-implement`, `/aif-improve`) reproduce the same stem at lookup
  time, so producer and consumers agree by construction.
- `NNNN = max(existing 4-digit prefix in paths.plans/) + 1`, zero-padded.
- Empty plans dir → start at `0001`.
- **Numbering is derived from currently existing numbered plan files.**
  If the highest-numbered plan is deleted or moved out of the directory,
  that number may be reused on the next run — keep plans in place if you
  rely on stable cross-references. (This is intentional: the algorithm
  reads files only, no managed counter is introduced.)
- **Bounded at `9999`.** If a `9999_*.md` plan already exists, `/aif-plan`
  aborts before writing instead of allowing `10000_…`, so consumer globs
  (also 4-digit) cannot drift out of contract.
- The prefix lives only on the **plan file**. Branch names stay
  `<branch_prefix><slug>` (no number) so existing git tooling, CI, and PR
  conventions are unaffected.
- Force-disabled when `HANDOFF_BRANCH_PREPARED=1` — the Handoff branch
  contract requires the plan filename to equal the branch name.
- Fast plans (`paths.plan`) and fix plans (`paths.fix_plan`) are single
  files and ignore this setting.

## Skill changes

- **aif-plan**: Step 0 resolves `workflow.plan_id_format`; Step 1.2.a
  defines the canonical `plan_file_stem`; Step 1.2.c implements the
  4-digit numbering algorithm with the 9999 hard cap and `Glob`-only
  enumeration (no `Bash(ls *)`); Step 5, `--list`, Important Rules, and
  Plan File Handling document the numbered shape. `timestamp` / `uuid`
  are explicitly documented as reserved values that behave like `slug`.
- **aif-implement**: branch-based plan discovery globs the numbered
  pattern first and falls back to the un-prefixed name. Multiple matches
  → highest-numbered + `WARN [aif-implement] multiple sequential plans …`.
- **aif-improve**: same numbered-first resolver, plus a hard invariant —
  preserve the existing 4-digit prefix on rewrite; never renumber.
- **aif-explore / aif-verify / aif-rules-check**: prose updates so plan
  references acknowledge the numbered path variant.

## Docs

- `docs/configuration.md`, `docs/config-reference.md`, `docs/plan-files.md`,
  `docs/workflow.md`, `docs/skills.md`, `AGENTS.md`,
  `skills/aif-plan/references/EXAMPLES.md`,
  `skills/aif-plan/references/TASK-FORMAT.md`.

## Test plan

- [ ] `npm test` — passes locally (94/94 on current head).
- [ ] Producer fixtures (assert exact branch-derived stem, production-equivalent allowed-tools — no `Bash(ls *)`):
  - [ ] `skills/aif-plan/tests/full-sequential-starts-at-0001.yaml` — empty plans dir, expects `0001_feature-add-product-search-api.md` and rejects the bare description slug.
  - [ ] `skills/aif-plan/tests/full-sequential-uses-next-number.yaml` — fixture with `0001/0002`, expects `0003_feature-add-product-search-api.md` and rejects the bare description slug.
  - [ ] `skills/aif-plan/tests/full-sequential-9999-cap.yaml` — fixture with `9999_alpha.md`, asserts no Write to `10000_*.md`, no overwrite of `9999_*.md`, no other 4-digit-prefixed plan, and output mentions the `9999` cap.
- [ ] Consumer fixture: `skills/aif-implement/tests/sequential-plan-discovery.yaml` — three plans (`0001_feature-user-auth.md`, `0005_feature-user-auth.md`, `0003_other-feature.md`); asserts `/aif-implement` reads `0005_*`, emits the documented WARN, ignores the older same-branch plan and the different-stem plan, and flips the checkbox only in `0005_*`.
- [ ] `skills/aif-plan/tests/fast-creates-plan-md.yaml` — unchanged; fast mode is independent of `plan_id_format`.
- [ ] Manual: run `/aif-plan full <description>` in a project with `workflow.plan_id_format: sequential` and a few existing `<NNNN>_*.md` plans → verify the next number lands and `/aif-implement` discovers it.
- [ ] Manual: empty `paths.plans/` → verify start at `0001`.
- [ ] Manual: `HANDOFF_BRANCH_PREPARED=1` flow → verify the prefix is suppressed.

## Backward compatibility

- Default `plan_id_format` remains `slug`. Existing projects see no behavior change.
- `timestamp` and `uuid` are reserved values that behave like `slug` (with an `INFO` log). They are not activated by this PR — activating them would need a separate filename-shape decision and matching consumer discovery.
- Unknown `plan_id_format` values fall back to `slug` with a `WARN [aif-plan]`.


ichinya:
This PR adds opt-in sequential full-plan filenames via `workflow.plan_id_format: sequential`, updates the main plan producer plus several consumers, and adds two scenario fixtures. The direction is good, but I would not merge/deploy this yet. There are still producer/consumer contract gaps that can make generated plans undiscoverable, plus a 9999+ rollover inconsistency and a test/production allowed-tools mismatch. CI also has not completed successfully; the only workflow run I found is `action_required`.

### What looks good

* The change keeps the default `slug` format unchanged and makes `sequential` opt-in.
* The PR correctly treats this as a cross-skill contract, not only an `/aif-plan` change: `/aif-implement`, `/aif-improve`, `/aif-verify`, `/aif-rules-check`, and `/aif-explore` all get some awareness of numbered plan paths.
* The Handoff prepared-branch override is called out, which is the right default because Handoff already owns branch identity.
* The two new `/aif-plan` fixtures cover the basic empty-directory and next-number cases.

### Must fix

1. [P1] Sequential plan producer can write a stem that the branch-based consumers will not find — `skills/aif-plan/SKILL.md`, `skills/aif-implement/SKILL.md`, `skills/aif-improve/SKILL.md`

   The PR summary and docs describe the file as `<NNNN>_<branch-or-slug>.md`, and the examples use branch-derived filenames such as `0004_feature-user-authentication.md`. But the actual Step 1.2 algorithm says:

   * generate a reusable slug from the description;
   * when sequential is enabled, use plan filename stem `"<prefix>_<slug>"`.

   In normal full mode with `git.enabled=true`, `git.create_branches=true`, and `git.branch_prefix=feature/`, this can produce `0003_product-search-api.md` while the active branch is `feature/product-search-api`. `/aif-implement` and `/aif-improve` then glob for `0003_feature-product-search-api.md` because they resolve from the current branch slug. If there are already multiple plans in `paths.plans/`, the fallback “single named full plan” path will also not rescue this; the user gets an ambiguous/missing plan instead of the normal `/aif-plan full` → `/aif-implement` flow.

   Expected fix:

   * Define one canonical `plan_file_stem` before saving:

     * Handoff prepared branch: `HANDOFF_BRANCH_NAME` with `/` replaced by `-`.
     * Git branch creation enabled: generated branch name with `/` replaced by `-`.
     * No branch / git disabled / branch creation disabled: description slug.
   * Apply the sequential prefix to that canonical stem, not to a separate bare description slug.
   * Update the two new fixtures to assert the exact branch-derived stem, not only `000N_[a-z0-9-]+.md`.
   * Add a consumer regression where `/aif-plan full` creates `0003_feature-product-search-api.md`, then `/aif-implement --list` or normal discovery on branch `feature/product-search-api` finds that exact file with multiple other plans present.

2. [P1] The 9999+ contract is internally inconsistent and breaks monotonic numbering/discovery — all sequential glob/regex call sites

   The new docs say the 4 digits are a minimum width and that after `9999_*.md` the format should continue as `10000_…`. But the algorithm and consumers only match exactly four digits:

   * producer scan: `^([0-9]{4})_.*\.md$`
   * producer glob: `[0-9][0-9][0-9][0-9]_*.md`
   * consumer globs: `[0-9][0-9][0-9][0-9]_<branch-slug>.md`
   * improve invariant: `^[0-9]{4}_.*\.md$`

   Scenario: once `10000_feature-a.md` exists, the next `/aif-plan full` ignores it, still sees `9999_*` as max, and reuses `10000`. Consumers also cannot discover `10000_<branch>.md` because they only glob four digits. That violates the stated monotonic contract and can lead to duplicate numbers or overwrite risk if the stem repeats.

   Expected fix:

   * Either explicitly cap at `9999` and emit a blocker/error before writing `10000_…`, or support variable-width numeric prefixes everywhere.
   * If supporting variable width, use a numeric-prefix parser like `^([0-9]+)_.*\.md$`, sort by parsed integer, and zero-pad only as `String(next).padStart(4, "0")` semantics in prose.
   * Update all consumers to find variable-width prefixes for the current branch stem.
   * Add regression coverage for:

     * existing `9999_alpha.md` → new file is `10000_<stem>.md`;
     * existing `10000_alpha.md` → next file is `10001_<stem>.md`;
     * `/aif-implement` and `/aif-improve` find `10000_<branch-stem>.md`.

3. [P1] Sequential algorithm recommends a Bash command that production `aif-plan` is not allowed to run, while tests override it — `skills/aif-plan/SKILL.md`, `skills/aif-plan/tests/*.yaml`

   The new algorithm says to use `Bash: ls <plans-dir> 2>/dev/null` or `Glob`. But the `aif-plan` frontmatter does not allow `Bash(ls *)`; it allows `Read`, `Write`, `Glob`, `Grep`, and a small set of `Bash(...)` patterns such as `git`, `cd`, `cp`, `mkdir`, and `basename`. The new fixtures add `Bash(ls *)` to `allowed_tools_override`, so tests can pass on a tool path that the installed production skill cannot use.

   That is a real runtime/test fidelity gap: an agent following the recommended `ls` path can hit a denied tool call in production, and the scenario tests would not catch it.

   Expected fix:

   * Prefer the safer fix: remove the `ls` option from the algorithm and require `Glob <plans-dir>/[0-9]*_*.md` plus regex filtering.
   * Remove `Bash(ls *)` from the new fixtures so the tests run under production-equivalent permissions.
   * If `ls` is intentionally required, add `Bash(ls *)` to `aif-plan` frontmatter and keep the fixtures aligned with that frontmatter.

4. [P1] `timestamp` / `uuid` are now treated as active values without a matching cross-skill resolver — `skills/aif-plan/SKILL.md`, `skills/aif-implement/SKILL.md`, `skills/aif-improve/SKILL.md`

   Before this PR, `workflow.plan_id_format` was documented as reserved / not read by dedicated built-ins. This PR changes `/aif-plan` to read `workflow.plan_id_format` and lists `timestamp` and `uuid` as allowed values, then Step 5 says full mode should “use the format-appropriate stem” for them. But the PR does not define deterministic timestamp/uuid generation, and the branch-based consumers only add special handling for `sequential`; they do not resolve `<timestamp>_<branch-stem>.md` or `<uuid>_<branch-stem>.md`.

   Scenario: an existing project already has `workflow.plan_id_format: timestamp` from the config template. After this PR, `/aif-plan full` may start writing timestamp-prefixed plans, while `/aif-implement` on the branch still only looks for the unprefixed branch file or the sequential four-digit pattern. With more than one full plan present, discovery becomes ambiguous.

   Expected fix:

   * If `timestamp` and `uuid` are still reserved/no-op, explicitly treat them as `slug` for now and remove “format-appropriate stem” wording from `/aif-plan`.
   * If they are intended to become active now, specify their exact filename shapes and implement matching branch-based discovery in `/aif-implement`, `/aif-improve`, `/aif-verify`, and `/aif-rules-check`.
   * Add regression coverage for at least one existing `timestamp` or `uuid` config where `/aif-plan full` creates a file and `/aif-implement` finds it with multiple plans present.

### Should fix

1. [P2] Add consumer-side scenario fixtures, not only producer fixtures

   The PR changes plan discovery behavior in `/aif-implement`, `/aif-improve`, `/aif-verify`, and `/aif-rules-check`, but the added tests only assert that `/aif-plan` writes numbered files. Add at least one scenario for `/aif-implement` resolving the highest-numbered branch match, and one for duplicate branch matches producing the documented WARN while choosing the highest number.

2. [P2] Update the broader workflow docs that still describe only unnumbered full plans

   `docs/configuration.md`, `docs/config-reference.md`, and `docs/plan-files.md` were updated, but this is also a workflow behavior change. Repository guidance says workflow logic changes should be reflected in `docs/workflow.md` and `docs/skills.md`, and `AGENTS.md` still documents the full-plan path as only `.ai-factory/plans/<branch-or-slug>.md`. Add the sequential variant there or explicitly link to `docs/plan-files.md`.

3. [P2] CI must be rerun and complete successfully

   The workflow run I found for head `15eb0159cade6e9d8d05f4e645b70def8f4aa9c6` is `CI` with conclusion `action_required`, and I did not find successful commit statuses. Do not merge until the workflow is approved/rerun and `npm test` passes on this head.

### Nits

None.

### Context gates

* Architecture: error — the producer/consumer filename contract is not yet consistent for branch mode and 9999+ rollover.
* Package CHECKLIST: warn — I did not find a package-specific CHECKLIST for the touched skill packages; AGENTS/docs update rules do apply and are not fully satisfied.
* Docs sync: warn — config and plan-file docs were updated, but broader workflow references remain stale.
* API/web/client sync: pass — no API, web, generated client, or schema changes.
* Feature flag / risk gating: error — the config flag is opt-in and default-safe, but `timestamp`/`uuid` semantics and sequential production permissions are not fully gated.
* Migration / backward compatibility: error — existing projects with `timestamp`/`uuid` config can see ambiguous new behavior; sequential also breaks after `9999`.
* PR size / decomposition: pass — 14 files, 382 additions, 23 deletions; coherent scope and no hard size policy found.
* Stack / merge order: pass — base is `2.x`, base SHA matches current `2.x`, PR is mergeable, and I found no linked stack dependency or prior review thread.
* CI: error — CI is not green; current run is `action_required`.

### Verdict

Verdict: REQUEST_CHANGES until sequential producer/consumer naming, 9999+ rollover, production allowed-tool coverage, timestamp/uuid semantics, and CI are fixed.

ichinya (changes requested):
until sequential producer/consumer naming, 9999+ rollover, production allowed-tool coverage, timestamp/uuid semantics, and CI are fixed.

TimurTurdyev:
Thanks for the detailed review, @ichinya. Pushed 4 follow-up commits on top of the original 5 to address every blocker. Mapping below.

### P1 — must fix

1. **Producer/consumer stem mismatch** → `1b90c52 fix(aif-plan): pin sequential plan contract …`
   - `/aif-plan` Step 1.2 now defines a single canonical `plan_file_stem` *before* applying any prefix:
     1. `HANDOFF_BRANCH_PREPARED=1` → `HANDOFF_BRANCH_NAME` (slashes → `-`)
     2. `git.enabled && git.create_branches` → generated branch name (slashes → `-`)
     3. otherwise → description slug
   - The `NNNN_` prefix is attached to that canonical stem, not to a separate bare slug. Branch-based consumers (`/aif-implement`, `/aif-improve`) already reproduce the same stem at lookup time, so producer and consumers now agree by construction.

2. **9999+ rollover** → `1b90c52` (same commit)
   - Step 1.2.c is explicitly bounded: when the existing max prefix is `9999`, the algorithm aborts with a clear error and tells the user to archive or switch to `slug`. The 4-digit consumer globs stay aligned because the producer never writes a 5-digit prefix.

3. **`Bash(ls *)` not in production allowlist** → `1b90c52` + `dbcbaad test(aif-plan): assert exact branch-derived stem …`
   - The `ls` recommendation is removed from `aif-plan/SKILL.md`. Enumeration is `Glob` only.
   - `Bash(ls *)` is removed from both `full-sequential-*.yaml` fixtures so the tests now run with production-equivalent permissions.

4. **`timestamp` / `uuid` half-active** → `1b90c52` + docs in `7892c07`
   - Step 0, Step 5, the config template, and the docs now explicitly call out `timestamp` and `uuid` as reserved values that currently fall back to `slug` (with an `INFO` log). The earlier "format-appropriate stem" wording is removed.

### P2 — should fix

5. **Consumer-side fixture** → `90e4f74 test(aif-implement): cover sequential plan discovery …`
   - New fixture `skills/aif-implement/tests/sequential-plan-discovery.yaml` puts three plans into `paths.plans/`:
     - `0001_feature-user-auth.md` (older same-branch plan)
     - `0005_feature-user-auth.md` (highest-numbered same-branch plan — chosen)
     - `0003_other-feature.md` (different stem — must be ignored)
   - Asserts: `/aif-implement` reads `0005_*`, writes only its task output, does **not** execute the older or unrelated plan, emits the documented `WARN [aif-implement] multiple sequential plans …`, and flips the checkbox in `0005_*` only.

6. **Broader workflow docs** → `7892c07 docs: sync sequential plan numbering …`
   - `docs/workflow.md`, `docs/skills.md`, and `AGENTS.md` now mention the sequential variant alongside the default `<branch-or-slug>.md` form and link back to `docs/plan-files.md` for the contract.

7. **Tighter fixture assertions** → `dbcbaad` (same commit as #3)
   - Both sequential fixtures now assert the exact branch-derived stem (`0001_feature-add-product-search-api.md` / `0003_feature-add-product-search-api.md`) and explicitly reject the bare description slug. This guards against regressions where the producer drifts back to using the description slug under branch mode.

### Test status

`npm test` locally: 93/93 passed. CI on the new head requires a maintainer approval — flagging that explicitly per your CI gate.

### What I deliberately did **not** do

- I did not implement variable-width numeric prefixes (your alternative for the 9999 rollover). The capped contract is simpler, fits the 0001..9999 budget that any sane single project will hit, and keeps all consumer globs honest. Happy to switch to variable-width if you prefer; that would touch every consumer glob.
- I did not implement `timestamp` / `uuid` as active formats. They stay reserved, behaving exactly like `slug` for now. Activating them is a separate follow-up that needs filename-shape decisions and matching consumer support.

ichinya:
The follow-up commits addressed most of the earlier contract issues: the producer now defines a canonical `plan_file_stem`, `timestamp` / `uuid` are explicitly reserved and treated as `slug`, the `ls` path was removed from `/aif-plan`, docs were broadened, and a consumer-side `/aif-implement` fixture was added. I still would not merge yet because the “monotonic / deleted plans never free numbers” contract is still not implemented by the current algorithm, and the branch needs a rebase + fresh CI.

### Response pass

* Addressed: Producer/consumer stem mismatch is fixed in `/aif-plan` Step 1.2.a by deriving a single canonical `plan_file_stem` from Handoff branch → generated git branch → description slug, then applying the sequential prefix to that stem.
* Addressed: `9999+` drift between producer and consumers is mostly fixed by making `9999` a hard cap instead of allowing `10000_…`.
* Addressed: `/aif-plan` no longer recommends `Bash(ls *)`; the new producer fixtures also removed `Bash(ls *)` from their allowed tools.
* Addressed: `timestamp` and `uuid` are no longer half-active; they are now documented as reserved values that behave like `slug`.
* Addressed: Broader workflow docs were updated (`AGENTS.md`, `docs/workflow.md`, `docs/skills.md`) and a consumer-side `skills/aif-implement/tests/sequential-plan-discovery.yaml` fixture was added.
* Still unresolved: CI is not green/complete for current head; the current CI workflow run is still `action_required`.
* New blocker found: The PR still claims true monotonic numbering and “deleted plans do not free their numbers”, but the algorithm only scans existing files under `paths.plans/`, so deleting or archiving the highest-numbered plan reuses that number.
* New blocker found: The branch is behind current `2.x` by 26 commits with overlapping docs/test-infra areas; rebase and rerun CI before merge.

### What looks good

* The canonical `plan_file_stem` design is the right fix for branch-based discovery: `/aif-plan` and `/aif-implement` now agree on `feature/foo` → `feature-foo`.
* The reserved-value treatment for `timestamp` / `uuid` is safer than introducing undocumented filename shapes.
* The new `/aif-implement` fixture tests the highest-numbered match and verifies that older or different-stem plans are not executed.
* The `9999` hard-cap decision keeps the four-digit consumer glob contract internally consistent.

### Must fix

1. [P1] “Monotonic / deleted plans never free numbers” is still false — `skills/aif-plan/SKILL.md`, `skills/aif/references/config-template.yaml`, `docs/config-reference.md`, `docs/plan-files.md`, PR body

   The sequential algorithm still derives `next` only from the currently existing files matched by:

   ```text
   <configured plans dir>/[0-9][0-9][0-9][0-9]_*.md
   ```

   That means the advertised contract is not true. Concrete scenario:

   1. `paths.plans/` contains:

      * `0001_alpha.md`
      * `0002_beta.md`
      * `0003_gamma.md`
   2. User deletes or archives `0003_gamma.md`.
   3. Next `/aif-plan full` with `workflow.plan_id_format: sequential` scans only existing files, sees max prefix `0002`, and writes another `0003_<stem>.md`.

   This directly contradicts the PR body and config template language: “Numbering is monotonic — deleted plans do not free their numbers.” It also conflicts with the current cap guidance: when max is `9999`, the error tells the user to “archive old plans”, but archiving old plans removes the only source of truth the algorithm uses, so it makes number reuse possible.

   Expected fix:

   * Choose one explicit contract:

     * **True monotonic:** persist the last allocated number in a managed counter/state file under `.ai-factory` or `paths.plans`, compute `next = max(counter, existing_max) + 1`, update the counter only after a successful write, and never lower it when files are deleted or archived.
     * **Existing-files-only:** remove “monotonic”, “deleted plans do not free their numbers”, and “archive old plans” wording from the PR body, config template, and docs. State that numbering is based on currently existing numbered plan files and that deleting the highest-numbered file can allow reuse.
   * Add regression coverage for the chosen contract:

     * true monotonic: counter says `0003`, files only contain `0001`/`0002`, next write must be `0004_<stem>.md`;
     * existing-files-only: docs/test description must explicitly allow reuse after highest-number deletion, and the feature should not claim monotonic uniqueness.

1. [P1] Do not merge/deploy this PR until it is rebased onto current `2.x` and CI is rerun

   The PR base SHA is 0d524b935681363801cc29aa55a1231027c3fe12, but current `2.x` is 3626f1914185cd2a0e60620cfe6125a963489e69. The branch is behind by 26 commits. The base moved through overlapping documentation/test-infra areas including `docs/configuration.md`, `docs/plan-files.md`, `docs/skills.md`, `scripts/test-skills.sh`, and package-lock changes, so stale-base confidence is low even though the PR currently reports mergeable.

   Acceptable options:

   * rebase onto current `2.x` and rerun CI;
   * if a rebase creates conflicts in docs/test fixtures, resolve them against the current base behavior and rerun the relevant skill tests;
   * link evidence if CI remains blocked by an unrelated GitHub Actions approval gate.

### Should fix

1. [P2] Update the PR body after the follow-up commits

   The PR body still describes the old test plan and still says “deleted plans never free their numbers.” It also does not mention the new `/aif-implement` consumer fixture or the reserved/no-op behavior for `timestamp` / `uuid` clearly enough. Once the contract decision above is made, update the body so reviewers and future maintainers do not have to infer behavior from the diff.

2. [P2] Add explicit coverage for the `9999` cap path

   The cap is now the chosen resolution for the 5-digit drift problem, but I did not see a fixture proving that existing `9999_*.md` causes `/aif-plan full` to stop before writing. Add a scenario with `9999_alpha.md` and assert no `Write` call under `paths.plans/10000_*.md` or `paths.plans/9999_*.md`, plus an output error containing the cap message.

### Nits

None.

### Context gates

* Architecture: warn — producer/consumer stem contract is now coherent, but the advertised monotonic numbering contract is still not implemented.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these skill/doc files.
* Docs / skill contract sync: error — docs and PR body still claim deletion-proof monotonic numbering that the skill cannot enforce.
* API/web/client sync: n/a — no API, web, generated client, or schema changes.
* Security surfaces: n/a — no new security-sensitive surface.
* Feature flag / risk gating: pass — `workflow.plan_id_format` remains opt-in and defaults to `slug`.
* Env / Turbo / process plumbing: n/a — no env/Turbo plumbing changes.
* Managed assets / ownership: warn — if a persistent counter is added for true monotonic behavior, ownership and migration behavior must be documented; current PR avoids new managed state but overclaims behavior.
* Migration / backward compatibility: warn — default behavior remains safe, but stale-base overlap and the false monotonic claim need resolution before release.
* Dependency / lockfile / production install: n/a — no dependency changes in this PR.
* PR size / decomposition: pass — 18 files / 615 additions is larger than before but still coherent; most growth is targeted docs and tests.
* Stack / merge order / base freshness: error — branch is behind current `2.x` by 26 commits and overlaps changed docs/test areas.
* CI / audit: error — current CI run for head 7892c07603117e3445e0f61847427bdcc72c433d is `action_required`; no successful check result was visible.

### Verification

* Checked: current PR metadata, current head SHA, prior review/requested-changes comment, author response, changed filenames, file patches for `/aif-plan`, `/aif-implement`, `/aif-improve`, producer fixtures, consumer fixture, config/docs updates, base freshness, and current CI workflow state.
* Ran locally: not run; I did not have a local checkout with this branch and did not execute `npm test`.
* CI: current visible CI workflow run is completed with conclusion `action_required`, and no jobs were available from the run.
* Not verified: actual scenario-runner execution for the new YAML fixtures, `npm test`, and post-rebase behavior against current `2.x`.

### Verdict

Verdict: REQUEST_CHANGES until the monotonic/deletion contract is either implemented or corrected in docs/body, and the PR is rebased onto current `2.x` with CI rerun.

TimurTurdyev:
Thanks for the second pass, @ichinya. Addressed both P1 items and the two P2 items. Mapping below.

### P1 — monotonic / deletion contract

Picked **option 2 from your review** (existing-files-only). The PR no longer claims monotonic numbering or that deleted plans never free their numbers — that was an over-promise relative to the actual algorithm.

- `20bc6fa docs(aif-plan): correct sequential numbering contract to existing-files-only`
  - `skills/aif-plan/SKILL.md` Step 1.2.c and Plan File Handling now describe numbering as derived from currently existing numbered plans in `paths.plans/`. The Rules block now states: *"Deleting or moving a numbered plan out of the directory can free that number for reuse on the next run — keep plans in place if you rely on stable cross-references."*
  - The 9999 cap error no longer suggests archiving as a way to keep the contract; it explicitly notes that moving the 9999-numbered file out of the directory will free 9999 for the next plan to reuse.
  - `skills/aif/references/config-template.yaml`, `docs/config-reference.md`, `docs/plan-files.md`, `skills/aif-plan/references/EXAMPLES.md`, `skills/aif-plan/references/TASK-FORMAT.md` — all updated to match. `TASK-FORMAT.md` also corrects an earlier inconsistency where `timestamp` / `uuid` were still described as active formats; they are now consistently documented as reserved values that behave like `slug` (matching `SKILL.md` and `config-template.yaml`).

The PR body has been rewritten in place to match this contract — no leftover "monotonic" / "deleted plans never free their numbers" wording.

If you'd prefer a true monotonic contract instead, that's a different change (persistent counter under `.ai-factory/` or `paths.plans/`, plus migration semantics) and worth its own PR. For this PR I kept the simpler existing-files-only contract.

### P1 — rebase + CI

- Rebased `feature/sequential-plan-numbering` onto current `2.x` (`3626f19`). No conflicts: the 26 intervening commits touched adjacent files but not the same hunks (`scripts/test-skills.sh`, `docs/configuration.md`, `docs/plan-files.md`, `docs/skills.md` overlaps resolved cleanly during rebase application).
- New base SHA: `3626f1914185cd2a0e60620cfe6125a963489e69`. New head SHA: `6bfe463`.
- Force-pushed with `--force-with-lease` to `fork/feature/sequential-plan-numbering`. PR picked up the new head automatically.
- CI is still `action_required` on the new head — same first-time-contributor gate as the previous round. Flagging explicitly per your CI requirement; please approve the workflow when convenient.

### P2 — 9999 cap fixture

- `6bfe463 test(aif-plan): cover 9999 sequential cap path`
  - New fixture `skills/aif-plan/tests/full-sequential-9999-cap.yaml`:
    - Pre-populates `.ai-factory/plans/9999_alpha.md`.
    - Asserts `/aif-plan full` produces no `Write` call to `.ai-factory/plans/10000_*.md`.
    - Asserts no `Write` call to `.ai-factory/plans/9999_*.md` (no overwrite of the existing plan, no reclaim of `9999`).
    - Asserts no `Write` call to any other 4-digit-prefixed plan file (no silent fallback to a different number).
    - Asserts the output mentions `9999` (the cap message).

### P2 — PR body update

Rewritten in place. Now reflects:
- canonical `plan_file_stem` (Handoff → git branch → slug);
- existing-files-only numbering with explicit deletion semantics;
- 9999 hard cap;
- reserved-and-falls-back-to-slug status of `timestamp` / `uuid`;
- the producer + consumer fixture set (including the new 9999 cap fixture and the existing `/aif-implement` sequential-discovery fixture).

### Test status

`npm test` locally: **94/94 passed** on `6bfe463` (the previous run was 93/93 — the new 9999-cap fixture is the +1).

### What I deliberately did **not** do

- Did not introduce a persistent counter for true monotonic numbering — your review explicitly left this as an alternative, and the existing-files-only contract is simpler and avoids adding new managed state. Happy to split a separate PR for the counter approach if you'd prefer it instead.
- Did not activate `timestamp` / `uuid` — they remain reserved and behave like `slug`. Activating them is a separate filename-shape decision plus matching consumer discovery.

ichinya:
## Code review

Re-checked PR #108 at current head 6bfe463f8c7bd969ab38a3df607678fd75300894.

The follow-up commits materially improved the PR: the main `/aif-plan` producer contract is now much clearer, `/aif-implement` has consumer coverage, `timestamp`/`uuid` are no longer half-activated, the `9999` case is bounded, and CI is green on the current head. I would still not merge yet because the same branch-stem resolver was not carried through consistently to `/aif-verify`, `/aif-rules-check`, and `/aif-explore`.

### Response pass

* Addressed: producer/consumer stem mismatch for `/aif-plan` → `/aif-implement` / `/aif-improve`. `plan_file_stem` is now canonical and branch-derived when a branch is created.
* Addressed: `9999+` rollover inconsistency. The PR now chooses a strict `0001..9999` contract and aborts before `10000_…`.
* Addressed: production/test tool mismatch for `/aif-plan`; sequential enumeration is now `Glob`-only and the producer fixtures no longer grant `Bash(ls *)`.
* Addressed: `timestamp` / `uuid` are now explicitly reserved and fall back to `slug`.
* Addressed: broader docs and `AGENTS.md` were updated.
* Addressed: CI is now green for current head `6bfe463...`.
* Still unresolved: none of the original P1 blockers remain as originally reported.
* New blocker found: `/aif-verify`, `/aif-rules-check`, and `/aif-explore` still describe sequential branch lookup using raw `<branch>` instead of the slash-normalized branch stem.

### What looks good

* `/aif-plan` now defines one canonical `plan_file_stem` before applying `sequential`, which fixes the core producer/consumer mismatch.
* The `9999` hard cap is consistent with the four-digit consumer glob contract and has a dedicated fixture.
* `timestamp` and `uuid` are now documented as reserved/no-op values, avoiding a backward-compatibility trap for existing config files.
* The new `/aif-implement` fixture covers multiple same-branch sequential plans and verifies highest-numbered selection plus the warning path.
* CI completed successfully on the current head.

### Must fix

1. [P1] Sequential plan lookup is still inconsistent in non-implement consumers — `skills/aif-verify/SKILL.md`, `skills/aif-rules-check/SKILL.md`, `skills/aif-explore/SKILL.md`

   `/aif-plan` now writes branch-mode sequential plans using the canonical branch stem, for example:

   `feature/user-auth` → `.ai-factory/plans/0005_feature-user-auth.md`

   `/aif-implement` and `/aif-improve` correctly say to convert the current branch by replacing `/` with `-` before globbing. But the updated `/aif-verify`, `/aif-rules-check`, and `/aif-explore` text still says variants of:

   * `paths.plans/<NNNN>_<branch>.md`
   * `<configured plans dir>/[0-9][0-9][0-9][0-9]_<branch-name>.md`
   * `paths.plans/<current-branch>.md`

   without defining `branch_stem = current branch with "/" replaced by "-"`.

   Broken scenario:

   * User enables `workflow.plan_id_format: sequential`.
   * `/aif-plan full Add user auth` creates branch `feature/user-auth` and plan `.ai-factory/plans/0005_feature-user-auth.md`.
   * `/aif-implement` can find and execute it.
   * User then runs `/aif-verify` on branch `feature/user-auth`.
   * The verify instructions can cause lookup for `.ai-factory/plans/0005_feature/user-auth.md` or otherwise fail to match the actual file. With multiple plans present, fallback discovery can become ambiguous and verification no longer validates against the plan that was just implemented.

   This is a cross-skill contract bug in the standard plan → implement → verify flow.

   Expected fix:

   * In `/aif-verify`, `/aif-rules-check`, and `/aif-explore`, explicitly resolve `workflow.plan_id_format` during config load.
   * Use the same wording/algorithm as `/aif-implement`:

     * get current branch;
     * compute `branch_stem` by replacing every `/` with `-`;
     * when sequential is enabled, glob `<paths.plans>/[0-9][0-9][0-9][0-9]_<branch_stem>.md`;
     * pick the highest-numbered match before falling back to `<paths.plans>/<branch_stem>.md`.
   * Add at least one regression/smoke fixture for `/aif-verify` on branch `feature/user-auth` with:

     * `0001_feature-user-auth.md`;
     * `0005_feature-user-auth.md`;
     * `0003_other-feature.md`;
     * expected behavior: verify reads `0005_feature-user-auth.md`, not the older same-stem plan and not the other-stem plan.

### Should fix

1. [P2] Clarify where sequential numbering is allocated for `full --parallel`

   Sequential numbering is resolved in `/aif-plan` Step 1.2, while `full --parallel` later creates and `cd`s into a worktree before the plan is written. That leaves the numbering scope ambiguous: is `NNNN` derived from the source worktree’s `paths.plans`, the target worktree’s `paths.plans`, or a project-wide convention?

   This is not blocking the normal non-parallel flow, but it should be made explicit because otherwise parallel runs can compute a prefix from one directory and write the plan into another.

   Suggested fix: state whether numbering is source-worktree scoped or target-worktree scoped. If it is target-worktree scoped, compute the sequential prefix after the worktree is created and after `cd <WORKTREE>`. If it is source/global scoped, keep the plan write/copy behavior aligned with that source directory.

### Nits

None.

### Context gates

* Architecture: error — main producer and `/aif-implement` are now aligned, but `/aif-verify`, `/aif-rules-check`, and `/aif-explore` still do not reproduce the canonical branch-stem lookup.
* Package CHECKLIST: warn — no package-specific CHECKLIST found for these skill changes; general workflow/docs rules were mostly followed.
* Docs sync: warn — docs are much improved, but the remaining consumer resolver text is inconsistent.
* API/web/client sync: pass — no API, web, schema, generated client, or event contract changes.
* Feature flag / risk gating: warn — default remains safe and opt-in; remaining issue is cross-skill resolver consistency under the flag.
* Migration / backward compatibility: pass — `slug` remains default; `timestamp`/`uuid` now fall back to `slug`.
* PR size / decomposition: pass — 19 files, 740 additions, 52 deletions; larger than before but still one coherent workflow-contract change.
* Stack / merge order: pass — base is current `2.x`, PR is open, non-draft, mergeable, and I found no stack dependency.
* CI: pass — CI run `304` completed successfully on head `6bfe463...`.

### Verdict

Verdict: REQUEST_CHANGES until `/aif-verify`, `/aif-rules-check`, and `/aif-explore` use the same canonical branch-stem sequential plan resolver as `/aif-plan` and `/aif-implement`.

TimurTurdyev:
Thanks for the third pass, @ichinya. Addressed the P1 and the P2.

### P1 — canonical branch_stem in non-implement consumers

- `259cd02 fix(skills): align verify/rules-check/explore with canonical branch_stem resolver`
  - `/aif-verify` Step 0.2, `/aif-rules-check` Step 2.3, and `/aif-explore` plan-aware sections (load-context, plan-exists, and the explore→plan table) now compute the lookup stem the same way `/aif-plan`, `/aif-implement`, and `/aif-improve` do:
    1. `branch_stem = git branch --show-current` with every `/` replaced by `-` (so `feature/user-auth` → `feature-user-auth`).
    2. When `workflow.plan_id_format = sequential`, glob `<paths.plans>/[0-9][0-9][0-9][0-9]_<branch_stem>.md` first and pick the highest-numbered match; emit `WARN [<skill>] multiple sequential plans for <branch>: <list>; using <chosen>` when more than one matches.
    3. Fall back to `<paths.plans>/<branch_stem>.md`.
  - Plus a Step-0 fallback that matches `/aif-implement`: when the branch-based plan is missing or git mode is off, accept a single `*.md` full-mode plan in the plans dir (4-digit prefix counts as a match), then fast plan, then standalone choices.

- `50f597a test(aif-verify): cover sequential plan discovery`
  - New fixture `skills/aif-verify/tests/sequential-plan-discovery.yaml` mirrors the existing `/aif-implement` consumer fixture for the verify side.
  - Fixture state on branch `feature/user-auth`:
    - `.ai-factory/plans/0001_feature-user-auth.md` — older, must not be chosen
    - `.ai-factory/plans/0005_feature-user-auth.md` — highest-numbered, chosen
    - `.ai-factory/plans/0003_other-feature.md` — different stem, must be ignored
  - Asserts: `/aif-verify` reads `0005_feature-user-auth.md`; does not read `0003_other-feature.md`; emits the documented `WARN [aif-verify] multiple sequential plans …`; stays in sandbox.
  - `allowed_tools_override` matches verify's production frontmatter (read-only; no `Write`/`Edit`).

### P2 — `--parallel` numbering scope

- `259cd02` (same commit as P1) makes the scope explicit in two places:
  - `/aif-plan` Step 1.2.c, implementation notes: a new bullet labels the sequential prefix as **source-worktree scoped** — computed from `<configured plans dir>` in the source worktree (the repo where `/aif-plan` was invoked), **before** the optional `cd <WORKTREE>` in Step 1.4. The plan file is then written to the same relative `<configured plans dir>/<NNNN>_<plan_file_stem>.md` path inside the target worktree.
  - `/aif-plan` Step 1.4 (`--parallel` worktree creation): a short callout repeats the rule — the prefix is already locked in by Step 1.2.c, do NOT recompute it after `cd <WORKTREE>` (the target worktree's plans dir is typically empty and would re-allocate `0001`, breaking the cross-worktree numbering contract on merge).

I chose source-worktree scope (over target-worktree or "project-wide via persistent counter") because the source dir is the only one with a meaningful number history at the moment `--parallel` is invoked; target-scope guarantees collisions across parallel runs, and a project-wide counter is the separate "true monotonic" PR you offered in the previous round.

### Test status

`npm test` locally: **94/94 passed** on `50f597a` (validator/script count; the new verify YAML fixture is documented as the consumer regression and is shaped identically to the existing producer/consumer fixtures).

### What I deliberately did **not** do

- Did not add a separate `--parallel` scenario fixture for sequential numbering. The producer-side numbering is the same code path in source-worktree mode; the change is purely documentation in this PR. Happy to add a parallel-mode fixture if you'd prefer explicit coverage — that would need the scenario runner to support worktree fixtures, which the current fixtures do not exercise.
- Did not touch `/aif-implement` / `/aif-improve` — they already had the canonical branch-stem algorithm from the first response round.

CI on the new head: should be in the same first-time-contributor `action_required` state until approval; flagging again per your CI gate.

ichinya:
## Code review

Re-checked PR #108 at current head 50f597a945b004f2e931b790879e3c959ac6f1be.

The PR is much closer. The previous branch-stem resolver issue was mostly addressed: `/aif-verify`, `/aif-rules-check`, and `/aif-explore` now use `branch_stem = current branch with "/" replaced by "-"`, and a new `/aif-verify` sequential discovery fixture was added. The remaining blocker is narrower: several consumers now reference `workflow.plan_id_format` in their plan lookup logic, but their config-loading sections still do not resolve that key. Current head also does not have a successful CI run yet; the latest CI run is `action_required`.

### Response pass

* Addressed: `/aif-verify` now computes `branch_stem` before sequential globbing.
* Addressed: `/aif-rules-check` now documents the same canonical branch-stem lookup.
* Addressed: `/aif-explore` now documents branch-stem-based sequential plan lookup.
* Addressed: a `/aif-verify` sequential plan discovery fixture was added.
* Still unresolved: consumers that use `workflow.plan_id_format` do not consistently list it as a resolved config key.
* New blocker found: current head CI is not green; latest CI run is `action_required`.

### What looks good

* The main producer contract in `/aif-plan` remains coherent: canonical `plan_file_stem`, `Glob`-only enumeration, `9999` cap, and reserved `timestamp` / `uuid`.
* `/aif-implement` and `/aif-verify` now both have scenario fixtures covering highest-numbered same-branch sequential plan discovery.
* The branch-stem wording is now present in `/aif-verify`, `/aif-rules-check`, and `/aif-explore`, which fixes the raw-branch-path problem from the prior pass.
* Base branch `2.x` is current, and the PR is still open, non-draft, and mergeable.

### Must fix

1. [P1] Consumers use `workflow.plan_id_format` without resolving it from config — `skills/aif-improve/SKILL.md`, `skills/aif-verify/SKILL.md`, `skills/aif-rules-check/SKILL.md`, `skills/aif-explore/SKILL.md`

   These skills now contain conditional logic like “when `workflow.plan_id_format = sequential`”, but their “Load Config” sections still do not include `workflow.plan_id_format` in the keys to resolve.

   Current state:

   * `/aif-implement` explicitly resolves `workflow.plan_id_format`, so it is aligned.
   * `/aif-improve` uses `workflow.plan_id_format` during plan lookup, but Step 0 only resolves paths, language, and git settings.
   * `/aif-verify` uses `workflow.plan_id_format` during plan lookup, but Step 0.0 only resolves paths, verify mode, git, and rules.
   * `/aif-rules-check` uses `workflow.plan_id_format` in optional plan context, but Step 1 only resolves paths, language, git, and rules.
   * `/aif-explore` uses `workflow.plan_id_format` for active-plan context, but Step 0 only resolves paths and language.

   Broken scenario:

   * Project config has `workflow.plan_id_format: sequential`.
   * `/aif-plan full` writes `.ai-factory/plans/0005_feature-user-auth.md`.
   * `/aif-implement` can find it because it resolves the flag.
   * `/aif-improve`, `/aif-verify`, `/aif-rules-check`, or `/aif-explore` may not resolve the flag and may treat plan discovery as default `slug`, falling back to `.ai-factory/plans/feature-user-auth.md` or ambiguous full-plan discovery.

   Expected fix:

   * Add `workflow.plan_id_format` to the config-loading/resolution section of every skill that branches on it:

     * `skills/aif-improve/SKILL.md`
     * `skills/aif-verify/SKILL.md`
     * `skills/aif-rules-check/SKILL.md`
     * `skills/aif-explore/SKILL.md`
   * State the same default everywhere: `workflow.plan_id_format` defaults to `slug`.
   * Keep semantics aligned with `/aif-plan`: active values are `slug` and `sequential`; `timestamp` / `uuid` are reserved and behave like `slug`.
   * Include this in list/read-only variants where applicable, especially `/aif-improve --list`.
   * Regression coverage: keep the new `/aif-verify` fixture, and add at least one lightweight fixture or validation check that the consumer explicitly reads/resolves `workflow.plan_id_format` before sequential lookup.

### Should fix

1. [P2] Clarify sequential numbering scope for `full --parallel`

   Sequential prefix allocation is described in Step 1.2 before the `full --parallel` worktree is created and before the skill `cd`s into the worktree. That leaves one operational ambiguity: is numbering derived from the source worktree’s `paths.plans`, the destination worktree’s `paths.plans`, or a project-wide source directory?

   Suggested fix:

   * state whether numbering is source-worktree scoped or target-worktree scoped;
   * if target-worktree scoped, compute the sequential prefix after worktree creation and after `cd <WORKTREE>`;
   * if source scoped, state how the plan file is written/copied so the prefix and destination directory stay consistent.

### Nits

None.

### Context gates

* Architecture: error — sequential discovery now has the right branch-stem shape, but some consumers still use the config flag without resolving it.
* Package CHECKLIST: warn — no package-specific CHECKLIST found for these skill files; general workflow/docs gates apply.
* Docs sync: warn — docs are mostly aligned, but consumer config-loading contracts still need update.
* API/web/client sync: pass — no API, web, schema, generated client, or event contract changes.
* Feature flag / risk gating: error — `workflow.plan_id_format` is opt-in and default-safe, but not consistently resolved by all consumers that branch on it.
* Migration / backward compatibility: pass — default remains `slug`; `timestamp` / `uuid` are reserved and fall back to `slug`.
* PR size / decomposition: pass — 20 changed files, 927 additions, 59 deletions; large, but still one coherent workflow-contract change.
* Stack / merge order: pass — base is current `2.x`; no stacked dependency found.
* CI: warn — latest CI for current head is `action_required`, not successful.

### Verdict

Verdict: REQUEST_CHANGES until all consumers that use `workflow.plan_id_format` explicitly resolve it from config and CI completes successfully on current head.

TimurTurdyev:
## Response to fourth-pass review

Thanks for the detailed re-check. Two follow-up commits on top of `50f597a`, no rebase (base hasn't moved):

* `66dda99` fix(skills): resolve `workflow.plan_id_format` in every consumer that branches on it
* `c95719d` test(aif-improve): cover sequential plan discovery in `--list`

New head: `c95719d`. `npm test` locally: 94/94 passed.

### P1 — Consumers now explicitly resolve `workflow.plan_id_format` — addressed

Every consumer that branches on the flag now lists it in its Load Config section with the same shape:

* default `slug`;
* active values `slug` and `sequential`;
* `timestamp` and `uuid` are **reserved** and currently behave like `slug`;
* unknown values are treated as `slug`.

Concrete edits in `66dda99`:

| File | Section | Change |
| --- | --- | --- |
| `skills/aif-improve/SKILL.md` | Step 0 (Load Config) + Step 0.list | Added `workflow.plan_id_format` bullet; Step 0.list now also globs `[0-9][0-9][0-9][0-9]_<branch-slug>.md` when sequential (it previously only checked the slug path — the `--list` regression you flagged). |
| `skills/aif-verify/SKILL.md` | Step 0.0 (Load config.yaml) | Added bullet; default also recorded under "If config.yaml doesn't exist". |
| `skills/aif-rules-check/SKILL.md` | Step 1 (Load Config) | Added to both the resolved-keys list and the defaults block. |
| `skills/aif-explore/SKILL.md` | Step 0 (Load Config) | Added bullet; default also recorded under "If config.yaml doesn't exist". |
| `skills/aif-implement/SKILL.md` | Step 0 | Wording aligned with the other four (reserved-value + unknown-value handling spelled out the same way) — purely cosmetic for consistency, behaviour unchanged. |

Wording template used everywhere (verbatim, modulo Step references):

> `workflow.plan_id_format` (default: `slug`) — used by branch-based plan discovery.
> Active values: `slug` and `sequential`. When `sequential`, the resolver globs
> `<paths.plans>/[0-9]{4}_<branch_stem>.md` first and falls back to
> `<paths.plans>/<branch_stem>.md` only if no numbered match is found.
> `timestamp` and `uuid` are **reserved values** and currently behave like `slug`.
> Treat any unknown value as `slug`.

### P1 regression coverage — addressed

New fixture `skills/aif-improve/tests/sequential-plan-discovery.yaml` (`c95719d`), modelled on `skills/aif-{implement,verify}/tests/sequential-plan-discovery.yaml` and adapted for the `--list` read-only contract.

Fixture state — `feature/user-auth`, three plans:

* `0001_feature-user-auth.md` — same stem, older;
* `0005_feature-user-auth.md` — same stem, highest;
* `0003_other-feature.md` — different stem.

Assertions:

* `Read .ai-factory/config.yaml` — proves the consumer actually resolves the flag before sequential lookup;
* `--list` output contains `0005_feature-user-auth.md`;
* `--list` output does NOT mention `0003_other-feature` as a current-branch sequential match;
* no `Write` call and no `Edit` call (`--list` is read-only by contract);
* `no_path_escape`.

`allowed_tools_override` for the runner is production-equivalent and intentionally read-only — `Write`/`Edit` are omitted because any write call from `--list` mode would already be a contract bug. The `/aif-verify` fixture from the previous round uses the same shape.

### P2 — `full --parallel` numbering scope — clarified

Already source-worktree scoped in Step 1.2.c of `skills/aif-plan/SKILL.md` and reiterated in Step 1.4, but I restructured the note into an explicit TL;DR so the answer to your three questions is unambiguous from a single block:

> * **Where the prefix is computed:** the source worktree's `<configured plans dir>` (the repo where `/aif-plan` was invoked) — i.e. exactly here, in Step 1.2.c.
> * **When it is computed:** **before** the optional `cd <WORKTREE>` in Step 1.4.
> * **Where the plan file is written:** the same relative `<configured plans dir>/<NNNN>_<plan_file_stem>.md` path inside the target worktree, so the prefix and destination directory stay consistent.
> * **What you must NOT do:** never recompute the prefix from the target worktree's plans dir after `cd <WORKTREE>`. The target dir is typically empty and would re-allocate `0001` on every parallel run, breaking the cross-worktree numbering contract on merge.

### CI

`action_required` on `50f597a` and `c95719d` looks like the first-time-contributor approval gate (same gate as on every previous head — `6bfe463` cleared it as run #304 only after manual approval). Could you trigger it on `c95719d` when convenient?

### Files in this round

* `skills/aif-improve/SKILL.md`
* `skills/aif-verify/SKILL.md`
* `skills/aif-rules-check/SKILL.md`
* `skills/aif-explore/SKILL.md`
* `skills/aif-implement/SKILL.md`
* `skills/aif-plan/SKILL.md`
* `skills/aif-improve/tests/sequential-plan-discovery.yaml` (new)

ichinya:
## Code review

Re-checked PR #108 at current head c95719d5eeea95693e0038ea18077b325910cf17.

This PR now has the sequential plan-file contract in a deployable shape from a code-review standpoint. The earlier producer/consumer mismatch, `9999` rollover ambiguity, production/test tool mismatch, `timestamp`/`uuid` half-activation, missing consumer config resolution, and `full --parallel` numbering-scope ambiguity have all been addressed. CI is green on the current head. The only remaining caveat is base freshness: `2.x` has advanced by 2 commits after this PR’s base, so I would update/rebase the branch and rerun CI before merging.

### Response pass

* Addressed: `/aif-plan` now computes a canonical `plan_file_stem` before applying the sequential prefix.
* Addressed: sequential numbering is explicitly bounded to `0001..9999`; producer errors before `10000_…`.
* Addressed: `/aif-plan` uses `Glob` only; no `Bash(ls *)` production/test mismatch remains.
* Addressed: `timestamp` and `uuid` are reserved and fall back to `slug`.
* Addressed: `/aif-implement`, `/aif-improve`, `/aif-verify`, `/aif-rules-check`, and `/aif-explore` now resolve or document `workflow.plan_id_format` consistently enough for the sequential branch-stem lookup.
* Addressed: `/aif-verify` and `/aif-improve` got sequential discovery regression fixtures.
* Addressed: `full --parallel` now explicitly locks the sequential prefix from the source worktree and avoids recomputing it after `cd <WORKTREE>`.
* Still unresolved: no meaningful code blockers remain.
* New blocker found: none.

### What looks good

* The producer and consumers now agree on the canonical branch-stem filename: branch `feature/user-auth` maps to `feature-user-auth`, and sequential files resolve as `<NNNN>_feature-user-auth.md`.
* The opt-in behavior is backward-compatible: default `slug` behavior is unchanged, and reserved `timestamp` / `uuid` do not silently activate new filename shapes.
* The `9999` hard cap keeps the four-digit glob contract consistent across producer and consumers.
* Regression coverage now includes producer numbering, cap behavior, `/aif-implement` sequential discovery, `/aif-verify` sequential discovery, and `/aif-improve --list` sequential discovery.
* CI run `306` completed successfully on head `c95719d...`.

### Must fix

I do not have meaningful code blockers anymore.

### Nits

None.

### Context gates

* Architecture: pass — the sequential producer/consumer filename contract is now consistent across the relevant skills.
* Package CHECKLIST: warn — no package-specific CHECKLIST found for these skill files; general workflow/docs requirements were followed.
* Docs sync: pass — config docs, plan-file docs, workflow docs, skills docs, AGENTS.md, and skill references were updated.
* API/web/client sync: pass — no API, web, schema, generated client, or event contract changes.
* Feature flag / risk gating: pass — `workflow.plan_id_format` is opt-in, default-safe, and resolved by consumers that depend on it.
* Migration / backward compatibility: pass — `slug` remains default; `timestamp` / `uuid` are reserved and behave like `slug`.
* PR size / decomposition: warn — 21 changed files and 1111 additions is large, but still a coherent cross-skill workflow-contract change.
* Stack / merge order: warn — no stacked dependency found, but the branch is behind current `2.x` by 2 commits and should be updated before merge.
* CI: pass with freshness caveat — CI succeeded on current head `c95719d...`; rerun after updating onto current `2.x`.

### Verdict

Verdict: LGTM

ichinya (approved):

PR #112, opened 2026-05-16
feat(aif-commit): use active plan Commit Plan for multi-commit grouping

### LINKED ISSUE #111: feat(aif-commit): use active plan Commit Plan for multi-commit grouping
## Problem

`/aif-plan` already writes a `## Commit Plan` for plans with 5+ tasks, and `/aif-implement` treats commit checkpoints as part of the plan-backed workflow. However, when a user runs `/aif-commit` after completing the whole plan, `/aif-commit` only analyzes staged changes and may create one commit even when the active plan contains multiple planned commits.

## Expected

`/aif-commit` should optionally use the active plan’s `## Commit Plan` as the first-class grouping source.

## Proposed behavior

1. Read `.ai-factory/config.yaml`.
2. Resolve active plan using the same plan priority as `/aif-implement`:
   - `@<plan-file>`
   - branch-based full plan
   - single full plan in `paths.plans`
   - `paths.plan`
3. If active plan contains `## Commit Plan`, parse commit groups and suggested messages.
4. Compare staged files/hunks with planned groups.
5. Ask:
   - Follow Commit Plan
   - Commit everything together
   - Adjust grouping
6. If no Commit Plan exists, keep current behavior unchanged.
7. If files cannot be mapped to groups, stop and ask user to adjust grouping.

## Non-goals

- `/aif-plan` does not commit.
- `/aif-implement` does not replace `/aif-commit`.
- `/aif-done` does not create commits.
- OpenSpec-specific `openspec/changes/<change-id>/tasks.md` support belongs in extensions as an adapter layer.

## Acceptance

- No behavior change when no `## Commit Plan` exists.
- A plan with 2 commit groups can produce 2 commits.
- Unmapped staged files trigger a user question.
- Existing split-unrelated-staged-work behavior remains.

ichinya:
## Summary

- Update `aif-commit` so commit planning can use active plan commit groups when deciding whether to split changes.
- Document the plan-aware grouping behavior in the workflow, skills, and config reference docs.
- Add smoke test coverage, including Windows portability fixes for the related test scripts.
- Closed #111 

## Impact

This makes commit guidance better aligned with existing plan checkpoints and reduces accidental grouping drift when a plan already defines commit boundaries.

## Validation

Not run in this publish step, per the project rule to avoid tests unless explicitly requested.

### DIFF
--- a/skills/aif-commit/SKILL.md
+++ b/skills/aif-commit/SKILL.md
@@ -79,8 +79,10 @@ If any rule is violated — fix the output before presenting it to the user.
      - use staged hunk evidence from `git diff --cached` when a file may span multiple groups
      - task ranges and `Files:` hints are guidance, not executable instructions
    - If files cannot be mapped to groups, stop and ask the user to adjust grouping.
-   - Only use `git add <files>` when each planned group has a disjoint file set.
+   - Before using whole-file staging, compare grouped files with unstaged worktree paths from `git diff --name-only`.
+   - Only use `git add <files>` when each planned group has a disjoint file set and no grouped file appears in `git diff --name-only`.
    - When one file spans multiple planned groups, use hunk-level staging (`git add -p` or `git apply --cached`) for each group.
+   - If grouped files overlap unstaged worktree paths, preserve and apply the original cached patch per group (`git diff --cached` + `git apply --cached`), use hunk-level staging, or stop before changing staging.
    - If hunk-level staging cannot be applied confidently, stop before changing staging and ask the user to adjust grouping or commit everything together.
    - When a usable grouping exists, ask:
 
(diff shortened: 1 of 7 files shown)


skills/aif-commit/SKILL.md
```diff
@@ -48,15 +51,58 @@ If any rule is violated — fix the output before presenting it to the user.
...
+   - If no active plan resolves or the active plan has no `## Commit Plan`, keep current staged-diff behavior unchanged.
+   - Never modify the active plan from this command.
+
+3. **Use Commit Plan Grouping When Available**
+   - If active plan contains `## Commit Plan`, parse:
+     - commit group number/name
+     - task range, such as `after tasks 1-3` or `tasks 4-6`
+     - suggested conventional commit message
+   - Read the plan's `## Tasks` or `## Implementation Tasks` section to map task ranges to task descriptions and any `Files:` hints.
+   - Compare staged files/hunks with planned groups before changing staging:
```
lee-to, line 77: This introduces hunk-level grouping, but the actual split flow below still unstages everything and stages each group with `git add <files>` before committing. If two planned groups touch the same file, `git add <file>` will stage all hunks for that file, so the first commit can accidentally include changes intended for a later group. Please either make the multi-commit flow use hunk-level staging (`git add -p` / patch application) or explicitly stop and ask the user when one file spans multiple groups.

lee-to (changes requested):
Found one blocking issue: the new hunk-aware grouping contract can currently produce incorrect commits when multiple planned groups touch the same file.

skills/aif-commit/SKILL.md
ichinya: Addressed with a hybrid contract: `/aif-commit` now only uses `git add <files>` when planned commit groups have disjoint file sets. If one file spans multiple planned groups, the skill requires hunk-level staging (`git add -p` or `git apply --cached`) for each group; if that cannot be applied confidently, it stops before changing staging and asks the user to adjust grouping or commit everything together.

I also added a regression check in `scripts/test-skills.sh` for this same-file cross-group case and updated `docs/workflow.md` / `docs/skills.md`.

Verification: `npm test` => 102 passed, 0 failed.

skills/aif-commit/SKILL.md
```diff
@@ -48,15 +51,61 @@ If any rule is violated — fix the output before presenting it to the user.
...
+     - use staged file paths from `git diff --cached --name-only`
+     - use staged hunk evidence from `git diff --cached` when a file may span multiple groups
+     - task ranges and `Files:` hints are guidance, not executable instructions
+   - If files cannot be mapped to groups, stop and ask the user to adjust grouping.
+   - Only use `git add <files>` when each planned group has a disjoint file set.
```
lee-to: This still allows whole-file staging when groups have disjoint file sets. If a grouped file has extra unstaged WIP hunks, the later split flow does `git reset HEAD` and then `git add <files>`, which stages every worktree hunk in that file, not just the hunks that were staged when `/aif-commit` started. That can commit changes the user intentionally left unstaged. Please make the whole-file path conditional on `git diff --name-only` having no overlap with the grouped files, or preserve/apply the original cached patch per group; otherwise fall back to hunk-level staging or stop before changing staging.

lee-to (changes requested):
Found one remaining staging-safety issue. The original same-file planned-group case is addressed, and npm test passes locally (102 passed, 0 failed), but the whole-file staging path can still pull in unstaged WIP hunks.

skills/aif-commit/SKILL.md
ichinya: Addressed in `60fefa5`. The whole-file path is now conditional on both checks: planned groups must have disjoint file sets, and none of the grouped files may appear in `git diff --name-only`.

If grouped files overlap unstaged worktree paths, `/aif-commit` must preserve/apply the original cached patch per group (`git diff --cached` + `git apply --cached`), use hunk-level staging, or stop before changing staging. I also added a regression check in `scripts/test-skills.sh` for the unstaged-WIP overlap case and updated the docs.

Verification before push: `npm test` => 103 passed, 0 failed. GitHub CI is running on the pushed commit now.

ichinya:
Update on the remaining staging-safety review item:

Pushed `60fefa5` with the follow-up guard. Whole-file staging is now allowed only when planned groups have disjoint file sets and none of the grouped files appear in `git diff --name-only`. If grouped files overlap unstaged worktree paths, `/aif-commit` must preserve/apply the original cached patch per group, use hunk-level staging, or stop before changing staging.

Added regression coverage in `scripts/test-skills.sh` for the unstaged-WIP overlap case and updated `docs/workflow.md` / `docs/skills.md`.

Local verification before push: `npm test` => 103 passed, 0 failed. GitHub CI is currently running on `60fefa5`.

lee-to (approved):
Rechecked the latest fix in 60fefa5. The planned grouping flow now guards both same-file commit groups and grouped files with unstaged worktree overlap before using whole-file staging. Local npm test passes (103 passed, 0 failed), and the GitHub test check is green.

PR #114, opened 2026-05-19
Add full-no-git-switching mode to aif-plan

dealenx:
image

### Add full-no-git-switching mode to /aif-plan

New planning mode that creates a full plan file in paths.plans/ with the description slug (no branch name), but unconditionally skips git branch/worktree creation.

Changes:
- skills/aif-plan/SKILL.md — add full-no-git-switching keyword, no_git_switching internal flag that short-circuits Step 1.4 (branch setup) and uses slug in Step 1.2.a, interactive menu option
- AGENTS.md — update workflow diagram and skill description

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -39,7 +39,7 @@ ai-factory/
 │   ├── aif-implement/          # Execute plan tasks
 │   ├── aif-improve/            # Plan refinement (second iteration)
 │   ├── aif-loop/               # Iterative reflex loop with quality gates
-│   ├── aif-plan/               # Plan implementation (fast/full modes)
+│   ├── aif-plan/               # Plan implementation (fast/full/full-no-git-switching modes)
 │   ├── aif-review/             # Code review
 │   ├── aif-roadmap/            # Strategic project roadmap
 │   ├── aif-rules/              # Project rules and conventions
@@ -229,7 +229,7 @@ Subsequent → review progress, add/reprioritize/mark milestones done
     ↓
 ROADMAP.md = strategic checklist of high-level goals
 
-/aif-plan [fast|full] <description>
+/aif-plan [fast|full|full-no-git-switching] <description>
     ↓
 Reads .ai-factory/DESCRIPTION.md + ARCHITECTURE.md for context
     ↓
@@ -239,6 +239,9 @@ full → creates richer plan, asks: tests? logging? docs?
        saves to configured `paths.plans/<branch-or-slug>.md`
        (`<slug>.md` when git is disabled or branch creation is off;
         `<NNNN>_<branch-or-slug>.md` when `workflow.plan_id_format: sequential`)
+full-no-git-switching → same as full but never creates or switches git branches
+       always uses description slug as plan file stem
+       saves to configured `paths.plans/<slug>.md`
     ↓
 Explores codebase
     ↓
--- a/skills/aif-plan/SKILL.md
+++ b/skills/aif-plan/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: aif-plan
-description: Plan implementation for a feature or task. Two modes — fast (single quick plan) or full (richer plan with optional git branch/worktree flow). Use when user says "plan", "new feature", "start feature", "create tasks".
-argument-hint: "[fast | full] [--parallel | --list | --cleanup <branch>] <description>"
+description: Plan implementation for a feature or task. Three modes — fast (single quick plan), full (richer plan with optional git branch/worktree flow), or full-no-git-switching (full plan without git branch switching). Use when user says "plan", "new feature", "start feature", "create tasks".
+argument-hint: "[fast | full | full-no-git-switching] [--parallel | --list | --cleanup <branch>] <description>"
 allowed-tools: Read Write Glob Grep Bash(git *) Bash(cd *) Bash(cp *) Bash(mkdir *) Bash(basename *) TaskCreate TaskUpdate TaskList AskUserQuestion Questions Task mcp__handoff__handoff_sync_status mcp__handoff__handoff_push_plan mcp__handoff__handoff_get_task mcp__handoff__handoff_list_tasks mcp__handoff__handoff_update_task
 disable-model-invocation: false
 version: 1.0.0
@@ -9,10 +9,11 @@ version: 1.0.0
 
 # Plan - Implementation Planning
 
-Create an implementation plan for a feature or task. Two modes:
+Create an implementation plan for a feature or task. Three modes:
 
 - **Fast** – quick plan, no git branch, saves to the configured fast plan path (default: `.ai-factory/PLAN.md`)
 - **Full** — richer plan, asks preferences, saves to the configured full-plan directory, and optionally creates a git branch/worktree when git is enabled and branch creation is allowed
+- **Full-no-git-switching** — identical to Full except it never creates or switches git branches; the plan file is saved to the plans/ directory using the description slug
 
 ## Workflow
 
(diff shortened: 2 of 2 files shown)


dealenx:
I did't make user tests of this fixed plan

dealenx:
Example: Using fixed `aif-plan`
image

ichinya:
## Code review

This PR adds a new public `/aif-plan full-no-git-switching` mode at head fdb766c9aaf727ecb0f1f295fb774c50e1388236. The intent is valid — users may want a full plan without branch/worktree switching — but the current diff is not safe to merge as-is because the new mode is documented but not actually wired into the skill’s branch-selection/runtime flow. There is also already an existing config-level mechanism for this behavior: `git.create_branches: false`.

### What looks good

* The PR is small and focused: only `AGENTS.md` and `skills/aif-plan/SKILL.md` are touched.
* The user-facing need is real: “full plan, but do not switch/create branches” is a useful workflow.
* The change keeps the plan-file location under `paths.plans/`, which is the right artifact family for full plans.

### Must fix

1. [P1] `full-no-git-switching` is declared but does not actually prevent branch/worktree creation — `skills/aif-plan/SKILL.md`

   The PR adds the new keyword to frontmatter, parsing text, and the interactive menu, but the operational steps still route through the existing full-mode branch logic.

   Broken scenario:

   ```text
   /aif-plan full-no-git-switching Add OAuth login
   ```

   With the default config:

   ```yaml
   git:
     enabled: true
     create_branches: true
   ```

   the current instructions still allow Step 1.2.a to choose the branch-backed case and Step 1.4 to run:

   ```bash
   git checkout <configured-base-branch>
   git pull origin <configured-base-branch>
   git checkout -b <branch-name>
   ```

   The PR body says it adds a `no_git_switching` internal flag that short-circuits Step 1.4, but the diff does not actually define or apply that flag. It also leaves “Mode selection” with only `fast`, `full`, and “neither” paths, so the new mode is not a reliable runtime state.

   This matters because the advertised behavior is the exact opposite of what can happen in the default git-enabled config: the command can still create/switch branches.

   Expected fix:

   * Prefer the simpler config-based solution unless there is a strong need for a new public mode:

     ```yaml
     git:
       create_branches: false
     ```

     Then update `AGENTS.md` / docs to explain that `/aif-plan full ...` already creates a rich full plan without branch switching when `git.create_branches=false`.
   * If keeping the new CLI mode, wire it explicitly:

     * parse `full-no-git-switching` as an exact/longest token before `full`;
     * set a clear internal state, e.g. `mode=full` + `no_git_switching=true`;
     * include it in Mode selection, not only in the option text;
     * force Step 1.2.a to use the description slug / no branch name when `no_git_switching=true`;
     * make Step 1.4 skip all branch/worktree commands when `no_git_switching=true`;
     * reject `--parallel` in this mode because parallel mode requires branch/worktree creation;
     * define precedence with `HANDOFF_BRANCH_PREPARED=1` so the Handoff branch filename contract is not accidentally broken.
   * Add a regression/contract check. A useful one would assert:

     ```text
     /aif-plan full-no-git-switching Add OAuth login
     -> mode=full
     -> no_git_switching=true
     -> branch_name is unset
     -> plan_file_stem=<description-slug>
     -> Step 1.4 does not run git checkout / git pull / git checkout -b / git worktree add
     ```

     Also cover:

     ```text
     /aif-plan full-no-git-switching --parallel Add OAuth login
     -> rejected with a short explanation
     ```

### Should fix

1. [P2] Public docs/reference are not synchronized if this remains a new mode

   `AGENTS.md` and `skills/aif-plan/SKILL.md` are changed, but the public skill docs/examples still describe `/aif-plan [fast|full]` and the existing config-driven branch behavior. If this mode is kept, update the argument examples and workflow docs as part of the same contract change. If the config-based solution is used instead, document `git.create_branches: false` as the no-branch full-plan path and remove the new mode.

### Nits

None.

### Context gates

* Architecture: n/a — prompt/docs-only change.
* Package CHECKLIST: n/a — no package boundary touched.
* Docs / skill contract sync: error — new public mode is documented in some places but not fully wired or synchronized with examples/docs.
* API/web/client sync: n/a — no API/client changes.
* Security surfaces: n/a — no security surface touched.
* Feature flag / risk gating: n/a — not a feature-flagged runtime change.
* Env / Turbo / process plumbing: n/a — no env/process changes.
* Managed assets / ownership: n/a — no managed asset install/update path touched.
* Migration / backward compatibility: warn — adding a public mode is not backward-breaking, but the mode currently has misleading behavior.
* Dependency / lockfile / production install: n/a — no dependency or lockfile changes.
* PR size / decomposition: pass — 2 files, small diff, single concern.
* Stack / merge order / base freshness: pass — PR is 1 commit ahead of `2.x` and not behind the current base from the inspected compare result.
* CI / audit: pass with source-review caveat — CI completed successfully, including build, lint, skill validation, and smoke tests; no dependency/audit surface changed.

### Verification

* Checked: PR metadata, changed files, full relevant `skills/aif-plan/SKILL.md` flow, `AGENTS.md` diff, config reference/template for `git.create_branches`, examples/docs sync, review threads, reviews, base freshness, and CI workflow state.
* Ran locally: not run; source review only.
* CI: current head has a successful `CI` workflow run.
* Not verified: real interactive Claude execution of `/aif-plan full-no-git-switching`; the PR author also noted that user testing was not performed.

Review brief followed from the uploaded reviewer instructions. 

### Verdict

Verdict: REQUEST_CHANGES until the new mode is either removed in favor of the existing `git.create_branches: false` config path, or fully wired so it cannot create/switch branches and has regression coverage.

lee-to:
Additional review, avoiding overlap with the existing review by @ichinya. I'm not repeating the already-covered runtime wiring / Step 1.4 / `--parallel` / public docs sync issues.

## New findings

1. **[P2] `full-no-git-switching` conflicts with the existing `workflow.plan_id_format: sequential` filename contract**

   References: `AGENTS.md:242-244`, `skills/aif-plan/SKILL.md:16`

   The new mode is described as always saving `paths.plans/<slug>.md`, but full-mode plans already support `workflow.plan_id_format: sequential`, where the actual filename must be `paths.plans/<NNNN>_<stem>.md`. If the new wording stays as-is, agents can reasonably skip the sequential prefix only for this mode. The safer wording is: `plan_file_stem=<description-slug>`, then still apply the normal Step 1.2.b / 1.2.c filename formatting.

2. **[P2] The research-default interactive path still only offers `full` vs `fast`**

   Reference: `skills/aif-plan/SKILL.md:223-225`

   After adding the third interactive option, the special case for an empty description with an `Active Summary -> Topic:` still says to “Only ask for `full` vs `fast`”. In that flow, users cannot pick `full-no-git-switching` even though the normal interactive menu exposes it. This should be updated to offer `full`, `full-no-git-switching`, and `fast`.

Source review only; I did not run the skill interactively.

dealenx:
Thanks for the thorough review, @ichinya and @lee-to. I agree the new mode isn't fully wired and needs proper integration with all the config contracts you've called out (plan_id_format: sequential, research-default interactive path, --parallel precedence, etc.).
At this point I'm not sure how to properly test this skill against all the configuration nuances and edge cases. I'll need some time to investigate before updating the PR. I'll explore both the config-based approach (git.create_branches: false) and the new CLI mode option, and come back with a properly wired solution that addresses all the raised concerns.

dealenx:
image

is `Full No-Git-Switch` mode supported in  v2.15?  Opencode offered 4 options for planning

lee-to:
> image
> is `Full No-Git-Switch` mode supported in v2.15? Opencode offered 4 options for planning

To enable full mode without a branch, you need to specify this in the config (.ai-factory/config.yaml)

PR #119, opened 2026-05-22
fix(aif): clean skills-lock.json on blocked-skill removal (fixes #117)

### LINKED ISSUE #117: Bug: /aif security-scan removal uses rm -rf - leaves skills-lock.json and CLI metadata stale
**Context:** SKILL.md for `aif` instructs (line 52) the agent to handle a
BLOCKED Level-1 security scan as:

> Exit 1 (BLOCKED) → Remove immediately (`rm -rf <skill-path>`),
> warn user. NEVER use.

## Problem
`rm -rf` deletes the physical directory but does not
touch `skills-lock.json` (and may not touch `.ai-factory.json`
managed-skills metadata, depending on whether the entry was added).
A later `npx skills experimental_install` (or a teammate cloning the
repo) re-installs the very skill AIF blocked for security reasons.

## Reproduction
(with `asyrafhussin/agent-skills@php-best-practices`):
1. Start `/aif` setup on an empty/new project and choose to install
   `asyrafhussin/agent-skills@php-best-practices` from skills.sh
   (or any other skill that will trip Level-1 scan; the exact skill
   does not matter as long as scan returns Exit 1).
2. AIF runs `npx skills add asyrafhussin/agent-skills --skill php-best-practices --agent claude-code --yes`.
3. `skills-lock.json` now contains an entry for `php-best-practices`.
4. AIF runs `python aif-skill-generator/scripts/security-scan.py .claude/skills/php-best-practices` and gets Exit 1 (BLOCKED).
5. Following its own SKILL.md instruction, AIF runs `rm -rf .claude/skills/php-best-practices`.
6. **Actual:** the physical directory is gone, but `skills-lock.json` STILL contains the `php-best-practices` entry.
7. Anyone running `npx skills experimental_install` afterwards (CI, a teammate cloning the repo) will reinstall the security-blocked skill.

### Why this matters specifically here
The skills being removed this way are flagged as *security threats*. Resurrecting them via
`experimental_install` defeats the whole purpose of the scan gate. 

### Environment
Agent: claude-code
LLM model: Opus 4.7 1M
Ai-factory version: 2.12.0

**If the bug is confirmed, I’m ready to submit a PR for this**

Meacue:
﻿## Summary

Replaces `rm -rf <skill-path>` with a dedicated cleanup helper that
also clears the stale entry in `skills-lock.json` AND **physically
removes** the security-blocked skill directory itself. Without this,
the blocked skill could be resurrected by
`npx skills experimental_install` or by a teammate cloning the repo
after install.

Closes #117

### New: `skills/aif-skill-generator/scripts/cleanup-blocked-skill.py`
- Calls `npx skills remove -s <name> -y` with `cwd=<root>` so the
  upstream CLI inspects the same project's `skills-lock.json` that
  this helper patches. Kept as best-effort first pass.
- Atomically removes `skills["<name>"]` from `skills-lock.json`
  (tmp file + `os.replace`, preserves indent / trailing newline).
- **Verifies by re-reading the patched lock file** (not by parsing
  the ANSI-coloured output of `npx skills list`, which is brittle and
  breaks for spaced names).
- `--installed-path <path>` is the **authoritative physical-removal**
  target. When supplied, the helper itself runs `safe_remove_installed`
  with strict defense-in-depth checks (the helper is the guarantor;
  the upstream CLI is best-effort):
  - resolve absolute or `--root`-relative path; idempotent if already
    gone
  - validate the target before any lock mutation: project-root
    containment, direct child of a known project agent `skillsDir`,
    `SKILL.md` marker, and identity match via sanitized basename or
    `SKILL.md` frontmatter `name:`
  - reject symlink/junction prefixes that redirect before the installed
    path itself
  - allow the managed upstream install shape where the installed path is
    a symlink/junction to another known project agent `skillsDir` child;
    cleanup removes the canonical target and then the link/junction
    path without following it
  - reject unmanaged or escaping symlinks/junctions, including targets
    outside `--root` or outside known project agent `skillsDir` roots
  - reject non-directory targets and explicit `installed == root`
    (`--installed-path .`)
  - delete via `shutil.rmtree` with `onerror` that clears the read-only
    bit
  - Caller contract: pass the SAME path the agent fed to
    `security-scan.py`, not a path synthesized from the logical skill
    name — upstream `skills` sanitizes the on-disk directory name
    (`"Convex Best Practices"` → `convex-best-practices`).
- Hard-fails when `npx` is missing in non-dry-run mode (previously
  exited 0 silently).
- Deny-list `--skill` validation: accepts the upstream CLI surface
  including spaces; rejects path separators, wildcards, control chars,
  leading hyphen (and leading whitespace + hyphen), and reserved
  `.` / `..`.

### Why physical removal lives in the helper

Upstream `npx skills remove -s <name>` matches `<name>` against
installed directory basenames via lowercased equality, WITHOUT
applying `sanitizeName()` to the argument. So
`npx skills remove -s "Convex Best Practices"` does not match
`.claude/skills/convex-best-practices` and exits 0 silently. Previously
the helper detected the leftover and exited 1, but the security-blocked
skill remained physically installed and the broad `Bash(rm -rf *)`
fallback was already removed in earlier rounds. The new
`safe_remove_installed` step makes the helper the guarantor of physical
removal without re-introducing a broad shell permission.

### Updated skill prompts (dropped `Bash(rm -rf *)` allowance)
- `skills/aif/SKILL.md` — 3 BLOCKED branches use the helper with
  `--installed-path <installed-skill-path>`.
- `skills/aif-skill-generator/SKILL.md` — 2 BLOCKED branches updated.
- `skills/aif-skill-generator/references/SECURITY-SCANNING.md` — 1
  BLOCKED branch updated.
- `allowed-tools` in both SKILL.md headers replaces
  `Bash(rm -rf *)` with `Bash(python *cleanup-blocked-skill*)`.

### New: `scripts/test-cleanup-helper.sh` (wired into CI)
Focused regression suite, uses a fake `npx` on PATH (no live network /
no registry):
- Lock removal + sibling preservation (#117 core regression).
- Skill name with spaces (validation + lock surface + argv).
- Bad-name rejection (empty, wildcards, path separators, leading hyphen,
  leading-whitespace hyphen, control characters, reserved names).
- No-op + available-keys hint when entry absent.
- Invalid JSON.
- Missing-`npx` hard-fail.
- Subprocess `cwd` propagation matches `--root`.
- `--installed-path` absent / leftover-under-skills/ → exit 0 + dir
  removed / partial-failure-downgrade.
- Sanitized-leftover end-to-end: with the correct scanned path AND the
  new helper, a `convex-best-practices` leftover is physically removed
  AND the lock key cleared, even when upstream `npx skills remove` is a
  silent no-op.
- Safety checks: path traversal, source/docs skill preservation,
  missing `SKILL.md`, file-not-directory, sibling preservation,
  permission failure, `installed-path == root`, direct-child known agent
  `skillsDir` enforcement, identity mismatch rejection, extension
  runtime `skillsDir` support, and dry-run validation.
- Managed symlink/junction coverage: normal upstream symlinked installs
  are accepted and cleaned; escaping symlinks/junctions are rejected and
  preserved; the lock entry is preserved when managed cleanup fails.
- Drift/contract guards: built-in helper roots stay in sync with
  `src/core/agents.ts`, and skill docs must reuse the scanned path for
  `--installed-path` instead of synthesizing `{{skills_dir}}/<name>`.

### CI
- `npm run test:cleanup-helper` runs on every PR via
  `.github/workflows/ci.yml`.

## Test plan

- [x] Manual end-to-end repro of #117:
  - install external skill → `skills-lock.json` has the entry
  - run cleanup helper → directory gone, lock entry gone
  - `npx skills experimental_install -y` → reports
    "No project skills found", skill NOT resurrected
- [x] `npm run test:cleanup-helper` — focused helper suite passes locally
- [x] `npm run lint` — clean
- [x] `npm test` — green 
- [x] CI green on `b079c26104e1057b0177206dc304a6d32fddfd6d`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,3 +38,6 @@ jobs:
 
       - name: Update smoke tests
         run: npm run test:update
+
+      - name: Cleanup helper tests
+        run: npm run test:cleanup-helper
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "test:update": "node scripts/run-bash-test.mjs scripts/test-update.sh",
     "test:extension-schema": "node scripts/test-extension-schema.mjs",
     "test:extensions": "node scripts/run-bash-test.mjs scripts/test-extensions.sh",
+    "test:cleanup-helper": "node scripts/run-bash-test.mjs scripts/test-cleanup-helper.sh",
     "lint:unused": "tsc --noEmit --noUnusedLocals --noUnusedParameters",
     "lint:dead": "knip --reporter compact",
     "lint": "npm run lint:unused && npm run lint:dead",
--- a/scripts/test-cleanup-helper.sh
+++ b/scripts/test-cleanup-helper.sh
@@ -100,6 +100,11 @@ write_lock() {
     printf '%s\n' "$2" > "$1/skills-lock.json"
 }
 
+write_config() {
+    # write_config <tmpdir> <json>
+    printf '%s\n' "$2" > "$1/.ai-factory.json"
+}
+
 echo -e "\n${BOLD}=== cleanup-blocked-skill.py regression tests ===${NC}\n"
 
 # ─────────────────────────────────────────────
@@ -360,10 +365,8 @@ rm -rf "$TEST_TMPDIR"
 # that physical directory, not just report leftover.
 #
 # Two sub-cases below probe both contracts on the same fixture:
-#   (a) regression-value: the OLD synthesized-path contract is still
-#       a silent no-op (path didn't exist), so without the new prompt
-#       contract a security-blocked dir would still leak. This proves
-#       why the prompt contract matters — it's a documentation case.
+#   (a) the OLD synthesized-path contract is now rejected instead of
+#       being accepted as an idempotent absent-path cleanup.
 #   (b) the FIX: with the correct path AND the new helper, the
 #       sanitized directory is physically removed and the lock key
 #       is cleared. This is the reviewer's explicit ask.
@@ -376,11 +379,9 @@ write_lock "$TEST_TMPDIR" '{
 mkdir -p "$TEST_TMPDIR/.claude/skills/convex-best-practices"
 echo "# blocked skill" > "$TEST_TMPDIR/.claude/skills/convex-best-practices/SKILL.md"
 
-# (a) OLD broken contract: synthesize path from logical name -> silent no-op.
-#     Fake npx returns 1 (simulates the upstream no-match case). With the
-#     bogus unsanitized installed-path the helper sees "absent" and exits 0
-#     (idempotent on missing target), proving that without the new prompt
-#     contract the REAL sanitized directory would be left untouched.
+# (a) OLD broken contract: synthesize path from logical name. The helper
+#     should now reject this absent wrong path by identity before it can
+#     silently claim cleanup of the real sanitized directory.
 if FAKE_NPX_EXIT=1 "$PYTHON" "$HELPER" \
         --skill "Convex Best Practices" \
         --root "$TEST_TMPDIR" \
@@ -390,12 +391,17 @@ if FAKE_NPX_EXIT=1 "$PYTHON" "$HELPER" \
 else
     OLD_CONTRACT_EXIT=$?
 fi
-# Verify: sanitized dir IS still on disk (the silent-leak hazard).
-if [[ ! -d "$TEST_TMPDIR/.claude/skills/convex-best-practices" ]]; then
-    fail "sanitized leftover (a)" "old-contract path removed the wrong dir"
+# Verify: sanitized dir IS still on disk, and the old synthesized path
+# is no longer accepted as a clean no-op.
+if [[ $OLD_CONTRACT_EXIT -eq 0 ]] || \
+   [[ ! -d "$TEST_TMPDIR/.claude/skills/convex-best-practices" ]] || \
+   ! grep -q 'identity mismatch' "$TEST_TMPDIR/out-a"; then
+    fail "sanitized leftover (a)" \
+         "old-contract path was not rejected as expected; exit=$OLD_CONTRACT_EXIT out=$(cat "$TEST_TMPDIR/out-a")"
 fi
 
-# Re-add the lock entry that path (a) removed, so path (b) has work to do.
+# Re-add the lock entry defensively so path (b) has work to do if a
+# future refactor moves validation later.
 write_lock "$TEST_TMPDIR" '{
   "version": 1,
   "skills": {"Convex Best Practices": {"source": "x/y"}}
@@ -423,11 +429,11 @@ if [[ ! -e "$TEST_TMPDIR/.claude/skills/convex-best-practices" ]]; then
     DIR_GONE=1
 fi
 
-if [[ $OLD_CONTRACT_EXIT -eq 0 \
+if [[ $OLD_CONTRACT_EXIT -ne 0 \
    && $NEW_CONTRACT_EXIT -eq 0 \
    && $LOCK_CLEAR -eq 1 \
    && $DIR_GONE -eq 1 ]]; then
-    pass "sanitized leftover: old-contract silent no-op + new-contract removes both"
+    pass "sanitized leftover: old-contract rejected + new-contract removes both"
 else
     fail "sanitized leftover" \
          "old=$OLD_CONTRACT_EXIT new=$NEW_CONTRACT_EXIT lock_clear=$LOCK_CLEAR dir_gone=$DIR_GONE; out-a=$(cat "$TEST_TMPDIR/out-a"); out-b=$(cat "$TEST_TMPDIR/out-b")"
@@ -516,8 +522,8 @@ fi
 rm -rf "$ESCAPE_DIR"
 rm -rf "$TEST_TMPDIR"
 
-# N3: --installed-path inside --root but NOT under any `skills/` segment
-# → reject (check #6).
+# N3: --installed-path inside --root but NOT under a known agent skills
+# directory → reject.
 fresh_tmp
 write_lock "$TEST_TMPDIR" '{
   "version": 1,
@@ -530,7 +536,7 @@ if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
     fail "N3: not under skills/" "helper exited 0 despite non-skills location"
 else
     if [[ -e "$TEST_TMPDIR/random-dir/SKILL.md" ]] && \
-       grep -q "no 'skills' segment" "$TEST_TMPDIR/out"; then
+       grep -Eq "known project agent skills directory|no 'skills' segment" "$TEST_TMPDIR/out"; then
         pass "N3: non-skills location rejected (dir preserved)"
     else
         fail "N3: not under skills/" \
@@ -661,14 +667,171 @@ if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
    --installed-path . > "$TEST_TMPDIR/out" 2>&1; then
     fail "N8: installed-path == root" "helper exited 0 despite installed-path == root"
 else
-    # Both #5b "equals --root" and #6 "no 'skills' segment" are valid
-    # rejection reasons (resolved root has no 'skills' part). Accept either.
+    # Both #5b "equals --root" and later path-boundary checks are valid
+    # rejection reasons. Accept either so ordering can stay defensive.
     if [[ -d "$TEST_TMPDIR" ]] && [[ -f "$TEST_TMPDIR/SKILL.md" ]] && \
-       grep -Eq "equals --root|no 'skills' segment" "$TEST_TMPDIR/out"; then
+       grep -Eq "equals --root|known project agent skills directory|no 'skills' segment" "$TEST_TMPDIR/out"; then
         pass "N8: installed-path == root rejected (root preserved)"
     else
         fail "N8: installed-path == root" \
              "root tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
+    fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N9: root-level source skill path → reject and preserve.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"source-skill": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/skills/source-skill"
+printf '%s\n' '---' 'name: source-skill' '---' '# source' > "$TEST_TMPDIR/skills/source-skill/SKILL.md"
+if "$PYTHON" "$HELPER" --skill source-skill --root "$TEST_TMPDIR" \
+   --installed-path skills/source-skill > "$TEST_TMPDIR/out" 2>&1; then
+    fail "N9: root source skills/ rejected" "helper exited 0 despite source path"
+else
+    if [[ -f "$TEST_TMPDIR/skills/source-skill/SKILL.md" ]] && \
+       grep -q "known project agent skills directory" "$TEST_TMPDIR/out"; then
+        pass "N9: root source skills/ rejected (dir preserved)"
+    else
+        fail "N9: root source skills/" \
+             "dir tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
+    fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N10: docs/skills source path → reject and preserve.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"source-skill": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/docs/skills/source-skill"
+printf '%s\n' '---' 'name: source-skill' '---' '# docs source' > "$TEST_TMPDIR/docs/skills/source-skill/SKILL.md"
+if "$PYTHON" "$HELPER" --skill source-skill --root "$TEST_TMPDIR" \
+   --installed-path docs/skills/source-skill > "$TEST_TMPDIR/out" 2>&1; then
+    fail "N10: docs/skills source rejected" "helper exited 0 despite docs path"
+else
+    if [[ -f "$TEST_TMPDIR/docs/skills/source-skill/SKILL.md" ]] && \
+       grep -q "known project agent skills directory" "$TEST_TMPDIR/out"; then
+        pass "N10: docs/skills source rejected (dir preserved)"
+    else
+        fail "N10: docs/skills source" \
+             "dir tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
+    fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N11: canonical .claude/skills/foo path still works.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"foo": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/.claude/skills/foo"
+printf '%s\n' '---' 'name: foo' '---' '# foo' > "$TEST_TMPDIR/.claude/skills/foo/SKILL.md"
+if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
+   --installed-path .claude/skills/foo > "$TEST_TMPDIR/out" 2>&1; then
+    if [[ ! -e "$TEST_TMPDIR/.claude/skills/foo" ]]; then
+        pass "N11: canonical .claude skill accepted and removed"
+    else
+        fail "N11: canonical .claude skill" \
+             "helper exited 0 but dir remains: $(cat "$TEST_TMPDIR/out")"
+    fi
+else
+    fail "N11: canonical .claude skill" "helper exited non-zero: $(cat "$TEST_TMPDIR/out")"
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N12: mismatched --skill and installed basename/frontmatter → reject.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"foo": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/.claude/skills/bar"
+printf '%s\n' '---' 'name: bar' '---' '# bar' > "$TEST_TMPDIR/.claude/skills/bar/SKILL.md"
+if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
+   --installed-path .claude/skills/bar > "$TEST_TMPDIR/out" 2>&1; then
+    fail "N12: identity mismatch rejected" "helper exited 0 despite mismatched path"
+else
+    if [[ -f "$TEST_TMPDIR/.claude/skills/bar/SKILL.md" ]] && \
+       grep -q "identity mismatch" "$TEST_TMPDIR/out"; then
+        pass "N12: identity mismatch rejected (dir preserved)"
+    else
+        fail "N12: identity mismatch" \
+             "dir tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
+    fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N13: .cache/skills/foo is rejected when not an explicit agent skillsDir.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"foo": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/.cache/skills/foo"
+printf '%s\n' '---' 'name: foo' '---' '# cache' > "$TEST_TMPDIR/.cache/skills/foo/SKILL.md"
+if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
+   --installed-path .cache/skills/foo > "$TEST_TMPDIR/out" 2>&1; then
+    fail "N13: unregistered .cache skills root rejected" "helper exited 0 despite unregistered root"
+else
+    if [[ -f "$TEST_TMPDIR/.cache/skills/foo/SKILL.md" ]] && \
+       grep -q "known project agent skills directory" "$TEST_TMPDIR/out"; then
+        pass "N13: unregistered .cache skills root rejected (dir preserved)"
+    else
+        fail "N13: unregistered .cache skills root" \
+             "dir tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
+    fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N14: extension runtime skillsDir from .ai-factory.json is accepted.
+fresh_tmp
+write_config "$TEST_TMPDIR" '{
+  "agents": [
+    {"id": "my-agent", "skillsDir": ".my-agent/skills"}
+  ]
+}'
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"foo": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/.my-agent/skills/foo"
+printf '%s\n' '---' 'name: foo' '---' '# extension runtime' > "$TEST_TMPDIR/.my-agent/skills/foo/SKILL.md"
+if "$PYTHON" "$HELPER" --skill foo --root "$TEST_TMPDIR" \
+   --installed-path .my-agent/skills/foo > "$TEST_TMPDIR/out" 2>&1; then
+    if [[ ! -e "$TEST_TMPDIR/.my-agent/skills/foo" ]]; then
+        pass "N14: extension runtime skillsDir accepted and removed"
+    else
+        fail "N14: extension runtime skillsDir" \
+             "helper exited 0 but dir remains: $(cat "$TEST_TMPDIR/out")"
+    fi
+else
+    fail "N14: extension runtime skillsDir" "helper exited non-zero: $(cat "$TEST_TMPDIR/out")"
+fi
+rm -rf "$TEST_TMPDIR"
+
+# N15: dry-run still validates unsafe installed paths and preserves them.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"source-skill": {"source": "x/y"}}
+}'
+mkdir -p "$TEST_TMPDIR/skills/source-skill"
+printf '%s\n' '---' 'name: source-skill' '---' '# source' > "$TEST_TMPDIR/skills/source-skill/SKILL.md"
+if "$PYTHON" "$HELPER" --skill source-skill --root "$TEST_TMPDIR" \
+   --installed-path skills/source-skill --dry-run > "$TEST_TMPDIR/out" 2>&1; then
+    fail "N15: dry-run validates source path" "helper exited 0 despite unsafe dry-run path"
+else
+    if [[ -f "$TEST_TMPDIR/skills/source-skill/SKILL.md" ]] && \
+       grep -q "known project agent skills directory" "$TEST_TMPDIR/out"; then
+        pass "N15: dry-run validates source path (dir preserved)"
+    else
+        fail "N15: dry-run validates source path" \
+             "dir tampered or wrong error: $(cat "$TEST_TMPDIR/out")"
     fi
 fi
 rm -rf "$TEST_TMPDIR"
--- a/scripts/test-cleanup-helper.sh
+++ b/scripts/test-cleanup-helper.sh
@@ -50,8 +50,11 @@ cat > "$FAKE_BIN/npx" << 'EOF'
 #!/bin/sh
 # Fake npx for cleanup-helper tests:
 # - logs each argv element on its own line into $NPX_ARGV_LOG
+# - logs the current working directory (as its basename) so tests can
+#   assert that the helper passes cwd=<root> to subprocess.run
 # - returns 0 unconditionally
 printf '%s\n' "$@" >> "${NPX_ARGV_LOG:-/dev/null}"
+echo "cwd_basename=$(basename "$(pwd)")" >> "${NPX_ARGV_LOG:-/dev/null}"
 exit 0
 EOF
 chmod +x "$FAKE_BIN/npx"
@@ -237,6 +240,37 @@ else
     else
         fail "missing npx fails fast" "helper failed but mutated the lock anyway"
     fi
+fi
+rm -rf "$TEST_TMPDIR"
+
+# ─────────────────────────────────────────────
+# Test 7: subprocess cwd matches --root (P2, second-review regression)
+# ─────────────────────────────────────────────
+# Ensures `npx skills remove` runs with cwd=<root> so the upstream CLI
+# inspects the same project's skills-lock.json that this helper patches.
+# Without cwd propagation, --root could patch one project while npx
+# acts on a different one.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"some-skill": {"source": "x/y"}}
+}'
+if "$PYTHON" "$HELPER" --skill some-skill --root "$TEST_TMPDIR" > "$TEST_TMPDIR/out" 2>&1; then
+    if [[ $IS_WINDOWS -eq 1 ]]; then
+        # Fake npx doesn't execute on Windows (PATHEXT — see top of file).
+        # The cwd-propagation contract is still covered by code inspection
+        # and Linux CI; document the platform gap.
+        pass "subprocess cwd matches --root (skipped on Windows)"
+    else
+        expected_basename=$(basename "$TEST_TMPDIR")
+        if grep -q "cwd_basename=${expected_basename}" "$NPX_ARGV_LOG"; then
+            pass "subprocess cwd matches --root"
+        else
+            fail "subprocess cwd matches --root" "subprocess cwd != --root. log: $(cat "$NPX_ARGV_LOG")"
+        fi
+    fi
+else
+    fail "subprocess cwd matches --root" "helper exited non-zero: $(cat "$TEST_TMPDIR/out")"
 fi
 rm -rf "$TEST_TMPDIR"
 
--- a/scripts/test-cleanup-helper.sh
+++ b/scripts/test-cleanup-helper.sh
@@ -346,6 +346,100 @@ else
 fi
 rm -rf "$TEST_TMPDIR"
 
+# ─────────────────────────────────────────────
+# Test 11: sanitized-leftover regression (P1 from 3rd review round)
+# ─────────────────────────────────────────────
+# Upstream `skills` sanitizes the on-disk directory name:
+#   "Convex Best Practices" -> "convex-best-practices"
+# An older prompt contract synthesized --installed-path as
+# `{{skills_dir}}/<name>`, i.e. ".claude/skills/Convex Best Practices",
+# which points at a path that NEVER existed. With cli_failed=True from
+# a partial npx failure, the helper would then see "directory absent",
+# downgrade to exit 0, and silently leave the REAL blocked directory
+# (sanitized path) on disk.
+#
+# This test exercises both contracts on the same fixture and proves
+# the new "pass the actual scanned path" contract is the only one
+# that surfaces the leftover.
+fresh_tmp
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"Convex Best Practices": {"source": "x/y"}}
+}'
+# Real leftover at the SANITIZED upstream path:
+mkdir -p "$TEST_TMPDIR/.claude/skills/convex-best-practices"
+echo "leftover" > "$TEST_TMPDIR/.claude/skills/convex-best-practices/SKILL.md"
+
+# (a) OLD broken contract: synthesize path from logical name -> FALSE PASS.
+#     Fake npx returns 1 to engage the downgrade path. With the bogus
+#     unsanitized installed-path the helper sees "absent" and exits 0,
+#     proving the silent-leak hazard the new contract avoids.
+if FAKE_NPX_EXIT=1 "$PYTHON" "$HELPER" \
+        --skill "Convex Best Practices" \
+        --root "$TEST_TMPDIR" \
+        --installed-path ".claude/skills/Convex Best Practices" \
+        > "$TEST_TMPDIR/out-a" 2>&1; then
+    OLD_CONTRACT_EXIT=0
+else
+    OLD_CONTRACT_EXIT=$?
+fi
+
+# Re-add the lock entry that path (a) removed, so path (b) has work to do.
+write_lock "$TEST_TMPDIR" '{
+  "version": 1,
+  "skills": {"Convex Best Practices": {"source": "x/y"}}
+}'
+
+# (b) NEW correct contract: pass the actual sanitized scanned path.
+#     The helper must see leftover and exit 1.
+if FAKE_NPX_EXIT=1 "$PYTHON" "$HELPER" \
+        --skill "Convex Best Practices" \
+        --root "$TEST_TMPDIR" \
+        --installed-path ".claude/skills/convex-best-practices" \
+        > "$TEST_TMPDIR/out-b" 2>&1; then
+    NEW_CONTRACT_EXIT=0
+else
+    NEW_CONTRACT_EXIT=$?
+fi
+
+if [[ $OLD_CONTRACT_EXIT -eq 0 && $NEW_CONTRACT_EXIT -ne 0 ]] \
+   && grep -q 'still present' "$TEST_TMPDIR/out-b"; then
+    pass "sanitized leftover: old contract false-passes, new contract exits 1"
+else
+    fail "sanitized leftover" "old=$OLD_CONTRACT_EXIT new=$NEW_CONTRACT_EXIT; out-a=$(cat "$TEST_TMPDIR/out-a"); out-b=$(cat "$TEST_TMPDIR/out-b")"
+fi
+rm -rf "$TEST_TMPDIR"
+
+# ─────────────────────────────────────────────
+# Test 12: prompt-contract grep — skill docs must NOT synthesize the
+# installed path from {{skills_dir}}/<name>; they must reuse the same
+# path token passed to security-scan.py.
+# ─────────────────────────────────────────────
+PROMPT_FILES=(
+    "$ROOT_DIR/skills/aif/SKILL.md"
+    "$ROOT_DIR/skills/aif-skill-generator/SKILL.md"
+    "$ROOT_DIR/skills/aif-skill-generator/references/SECURITY-SCANNING.md"
+)
+BAD_LINES=()
+for f in "${PROMPT_FILES[@]}"; do
+    if [[ ! -f "$f" ]]; then
+        continue
+    fi
+    # Regression = the synthesized template appears IMMEDIATELY after
+    # --installed-path (i.e. it is the VALUE of the flag, not part of
+    # surrounding explanatory text that warns against the pattern).
+    while IFS= read -r line; do
+        if printf '%s' "$line" | grep -Eq -- '--installed-path[[:space:]]+\{\{skills_dir\}\}/<'; then
+            BAD_LINES+=("$f: $line")
+        fi
+    done < "$f"
+done
+if [[ ${#BAD_LINES[@]} -eq 0 ]]; then
+    pass "prompt-contract: no skill doc synthesizes --installed-path from {{skills_dir}}/<name>"
+else
+    fail "prompt-contract regression" "$(printf '%s\n' "${BAD_LINES[@]}")"
+fi
+
 # Restore PATH for any post-suite work
 export PATH="$ORIGINAL_PATH"
 if [[ -n "$ORIGINAL_PATHEXT" ]]; then
--- a/scripts/test-cleanup-helper.sh
+++ b/scripts/test-cleanup-helper.sh
@@ -108,7 +108,7 @@ write_config() {
 echo -e "\n${BOLD}=== cleanup-blocked-skill.py regression tests ===${NC}\n"
 
 # ─────────────────────────────────────────────
-# Test 1: removes target, preserves siblings (#117 regression)
+# Test 1: removes target lock entry, preserves siblings
 # ─────────────────────────────────────────────
 fresh_tmp
 write_lock "$TEST_TMPDIR" '{
@@ -131,7 +131,7 @@ fi
 rm -rf "$TEST_TMPDIR"
 
 # ─────────────────────────────────────────────
-# Test 2: skill name with spaces (P1.1 regression)
+# Test 2: skill name with spaces
 # ─────────────────────────────────────────────
 fresh_tmp
 write_lock "$TEST_TMPDIR" '{
@@ -151,9 +151,9 @@ if "$PYTHON" "$HELPER" --skill "Convex Best Practices" --root "$TEST_TMPDIR" > "
     if [[ $LOCK_OK -eq 0 ]]; then
         fail "skill name with spaces" "lock content unexpected: $(cat "$TEST_TMPDIR/skills-lock.json")"
     elif [[ $IS_WINDOWS -eq 1 ]]; then
-        # Lock-state proves the P1.1 fix (validation + patch_lock both
-        # accepted the spaced name); argv-passing assertion is unreliable
-        # on Windows (see PATHEXT note above).
+        # Lock-state proves validation + patch_lock accepted the spaced
+        # name; argv-passing assertion is unreliable on Windows (see
+        # PATHEXT note above).
         pass "skill name with spaces (lock; argv check skipped on Windows)"
     else
         # Fake npx logged each argv on its own line — verify the full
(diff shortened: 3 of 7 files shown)


ichinya:
## Code review

PR #119 at head 6acbbd7b844f8e281fabcfed853ff02c559d067e replaces the old BLOCKED-skill `rm -rf` cleanup path with a dedicated helper that calls `npx skills remove`, patches `skills-lock.json`, and updates `/aif`, `/aif-skill-generator`, and the security-scanning reference. The direction is correct: it targets the exact #117 security failure where a blocked external skill can be resurrected from a stale project lock entry. I would not merge yet: the helper currently rejects part of the valid upstream skill-name surface, and CI is not green/actionable for the current head.  

### What looks good

* The PR removes the broad `Bash(rm -rf *)` allowance from both affected skills and replaces it with a narrower helper invocation. That is a better permission shape for a security cleanup path.  
* The helper patches only `skills[<name>]` in `skills-lock.json`, writes through a temp file, and preserves the rest of the lock object instead of rewriting unrelated state. 
* The prompt/docs updates cover the main BLOCKED paths in `/aif`, `/aif-skill-generator`, and `SECURITY-SCANNING.md`, so the behavioral contract is mostly synchronized.   

### Must fix

1. [P1] Cleanup helper rejects valid skills CLI names — `skills/aif-skill-generator/scripts/cleanup-blocked-skill.py`

   The new helper requires `--skill` to match `^[A-Za-z0-9._-]+$`. That is narrower than the upstream `skills` CLI surface: upstream docs explicitly allow installing a skill with spaces, for example `--skill "Convex Best Practices"`, and upstream tests treat spaces as valid input that are sanitized into a directory-safe name.   

   The changed `/aif` instructions pass `<name>` / `<skill-name>` directly to the helper after a BLOCKED scan. If that name contains spaces, the helper exits `2` before it calls `npx skills remove` and before it removes the local lock entry. That leaves the blocked skill on the same resurrection path described in #117.   

   Expected fix:

   * Accept the same logical skill-name surface that `skills remove` and `skills-lock.json` can represent, at least plain spaces.
   * Reject only genuinely unsafe values: empty/whitespace-only names, path separators, control characters, and wildcard/all selectors unless intentionally supported.
   * If the helper expects a canonical installed key rather than the user/search name, make that explicit and derive/validate it against `skills-lock.json` or the installed directory before mutating anything.
   * Add a regression test with a local `skills-lock.json` containing a skill key with spaces, and assert the helper removes only that entry and passes the full name as one argv value to the fake `npx` process.

2. [P1] Current head does not have a green CI merge signal — checks/base freshness

   The PR body still has `npm test` and CI unchecked, and the current GitHub Actions run for this head is completed with `action_required`, with no jobs visible from the API. The branch is also 3 commits behind current `2.x`. I did not see overlapping changed files in those behind commits, but this is still not a safe merge gate for a security cleanup PR.

   Expected fix:

   * Update/rebase the branch onto current `2.x`, or explicitly confirm no update is required if GitHub’s merge queue handles it.
   * Rerun CI on the resulting head and make the required checks green.
   * If `action_required` is only a maintainer-approval gate for first-time workflow execution, approve/rerun it and link the successful run.

### Should fix

1. [P2] Make cleanup success semantics stricter when `npx` or verification is unavailable

   `run_skills_remove()` returns success when `npx` is not on `PATH`, and `verify_removed()` also returns success when it cannot run `npx`. In that path the helper can exit `0` after only patching the lock file, while the blocked skill directory may still exist. That weakens the PR body’s claim that the helper deletes the directory and clears the lock entry. 

   Expected fix:

   * Treat missing `npx` as an operational failure outside `--dry-run`, or add an explicit lock-only mode whose limitations are visible in the output.
   * Check the `npx skills list` return code; do not treat a failed list command as successful verification.
   * Add fake-`npx` tests for: remove succeeds/list absent, remove fails/list still shows skill, remove fails/list absent, and `npx` missing.

2. [P2] Add focused automated coverage for the helper

   The manual repro is useful, but this is a security regression path and the repo test plan is still unchecked. The minimal useful coverage is small and does not require live registry/network access.

   Expected fix:

   * Add a temp-directory test for lock patching that preserves unrelated keys and removes only the blocked skill.
   * Add invalid JSON and missing `skills` object cases.
   * Add a fake `npx` script so the test asserts subprocess argv and verification behavior without depending on the real CLI or network.
   * Include the valid-name-with-spaces case from the P1 blocker.

### Nits

None.

### Context gates

* Architecture: pass — the change is isolated to skill prompt contracts plus one helper script; no broader CLI/package boundary is touched.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths; I inspected the general repo guidance and quality-gate docs instead.  
* Docs / skill contract sync: error — the skills now instruct callers to use the helper, but the helper rejects valid upstream skill names.
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: error — this fixes the stale-lock resurrection path for simple names, but still fails before cleanup for valid names with spaces and can report success when CLI verification cannot run.
* Feature flag / risk gating: n/a — no runtime feature flag or rollout switch needed for this prompt/helper-only security fix.
* Env / Turbo / process plumbing: n/a — no env, Turbo, or process-boundary changes.
* Managed assets / ownership: warn — the helper intentionally mutates `skills-lock.json`, but should more strongly prove the installed blocked skill was actually removed before reporting success.
* Migration / backward compatibility: error — the new cleanup helper is narrower than the upstream CLI’s accepted skill-name contract.
* Dependency / lockfile / production install: pass — no npm dependency or package lock change.
* PR size / decomposition: pass — 4 files and a cohesive security fix.
* Stack / merge order / base freshness: warn — not stacked, but the branch is behind `2.x`; update/rebase and rerun CI before merge.
* CI / audit: error — current CI is not green/actionable for the head under review.

### Verification

* Checked: PR metadata, diff, changed skill docs, new helper script, linked issue #117, upstream `skills` CLI remove/list/local-lock behavior, repo package scripts, and the supplied review rubric.      
* Ran locally: `python3 -m py_compile` against the helper reconstructed from the PR patch; no syntax error.
* CI: current known state is not green; GitHub Actions `CI` is `completed/action_required` for the head, with no jobs visible through the API.
* Not verified: full `npm test`, live `npx skills` end-to-end behavior, registry/GitHub fetch behavior, and maintainer-only required checks/audit.

### Verdict

Verdict: REQUEST_CHANGES until the helper accepts the valid skills CLI name surface or safely derives the installed lock key, the security regression tests cover that path, and CI is rerun/green on the current head.

lee-to:
## Additional review

I checked the current diff against the existing review and found two additional issues that are not covered there.

1. **P1: One cleanup path uses a non-resolvable short script name**

   In `skills/aif/SKILL.md`, the Skill Acquisition Strategy block now says:

   ```bash
   $PYTHON cleanup-blocked-skill.py --skill <name>
   ```

   The helper is installed under `~/{{skills_dir}}/aif-skill-generator/scripts/cleanup-blocked-skill.py`, and the `/aif` workflow normally runs from the target project root. In that context the shortened command will usually fail with “file not found”, leaving the blocked skill and stale lock entry in place. The other updated call sites use the full helper path, so this one should be made consistent.

2. **P2: `--root` patches one project while `npx skills remove/list` runs in the current cwd**

   `cleanup-blocked-skill.py` accepts `--root` and patches `<root>/skills-lock.json`, but both subprocess calls run without `cwd=root`:

   ```python
   subprocess.run([npx, "skills", "remove", "-s", skill, "-y"])
   subprocess.run([npx, "skills", "list"], ...)
   ```

   If the helper is launched from a subdirectory or from another cwd with `--root /path/to/project`, it can patch the intended project lock file while removing/verifying skills in a different project context. Please run both `npx` calls with `cwd=str(root)` (threading `root` through `run_skills_remove()` and `verify_removed()`), or remove `--root` if cross-cwd use is intentionally unsupported.

I intentionally did not repeat the existing review points about skill names with spaces, weak success semantics when `npx` is unavailable/fails, missing focused tests, or CI/base freshness.

Meacue:
Thank you very much for reviewing the code, @ichinya and @lee-to. I will implement the suggested changes and address the identified issues shortly.

Meacue:
Applied fixes for both reviews across three commits. Each item below maps to the original review point so re-verification is straightforward.

### Commit `de5427f` - addresses @ichinya's review

- **P1** Skill-name validation loosened to a deny-list (`validate_skill_name`) that accepts the upstream CLI surface including spaces (e.g. `"Convex Best Practices"`). Rejects only genuinely unsafe values: path separators, wildcards, control chars, leading hyphen, reserved `.`/`..`, empty/whitespace-only.
- **P2.1** Hard-fail when `npx` is missing in non-dry-run mode (previously a silent `exit 0`). Replaced the fragile `verify_removed` (ANSI-parsed `npx skills list` output, broke on spaced names) with `verify_lock_clean` — deterministic re-read of the patched lock file.
- **P2.2** New `scripts/test-cleanup-helper.sh` with a fake-npx that logs argv; wired into `package.json` and `.github/workflows/ci.yml`. Covers #117 regression, spaced names, bad-name rejection, no-op + available-keys message, invalid JSON, and missing-npx fail-fast.

### Commit `cd48fc0` - addresses @lee-to's additional review

- **P1** The short-form `cleanup-blocked-skill.py` invocation in `skills/aif/SKILL.md:94` could not resolve from a project cwd; replaced with the full `~/{{skills_dir}}/aif-skill-generator/scripts/cleanup-blocked-skill.py` path so all call sites are consistent.
- **P2** `run_skills_remove()` now accepts `root: Path` and runs `subprocess.run(cmd, cwd=str(root))` for both `remove` and (back then) `list` calls, so `--root` and the upstream CLI agree on the project context. New Test 7 asserts cwd propagation via the fake-npx log.

### Commit `e507c7d` - follow-up from a self-audit

Four findings I surfaced after the two review rounds. Three required code; one was dropped after upstream-CLI investigation.

- **New `--installed-path <path>` flag** on the helper. When supplied, the helper verifies the physical skill directory after patching `skills-lock.json`:
  - dir still present → `exit 1` (also catches the previously-silent `rc=0` + leftover-directory leak)
  - dir is gone → downgrade partial-failure (`npx remove` returned non-zero) to `exit 0` since the security goal is met
  Behavior without `--installed-path` is unchanged for backward compat. Threaded through all 6 agent-facing call sites as `--installed-path {{skills_dir}}/<name>` (template resolves per agent via `src/core/template.ts`).
- **Leading-whitespace bypass closed** in `validate_skill_name`: `" -foo"` previously slipped past the leading-hyphen check; now uses `name.lstrip().startswith('-')`.
- **Test 6 hardened on Windows**: `shutil.which` may try extension fallbacks (`npx.cmd`) even with sparse PATH; test now runs with both `PATH=""` and `PATHEXT=""`. Original PATHEXT saved/restored in suite setup.
- **New tests 8/9/10** — `--installed-path` absent on disk → `exit 0`; `--installed-path` leftover → `exit 1` (uses a non-standard subdir so a real `npx` running on Windows via PATHEXT cannot delete the assertion artifact); `FAKE_NPX_EXIT=1` + dir gone → `exit 0` (downgrade). Fake-npx now honors `FAKE_NPX_EXIT`. Test 3 `BAD_INPUTS` extended with `" -foo"` (issue #3 regression) and `"   "`.
- **Intentionally not changed**: no explicit `--scope project` on `npx skills remove`. Upstream `vercel-labs/skills` defaults to project scope; only `-g` opts into global. The existing `cwd=str(root)` is the canonical way to target the right project, and adding a non-existent flag would break the call.

@ichinya @lee-to, could you please re-verify the response items above?

ichinya:
## Code review

Follow-up review for PR #119 at current head e507c7db2e4de4323bc0bae253cf5ab97870d3f5. The branch has moved since the previous review: it now includes the helper regression suite, CI wiring, looser skill-name validation, stricter `npx` failure handling, `cwd=root` subprocess behavior, and the new `--installed-path` verification path. The earlier implementation concerns are mostly addressed, but I still would not merge yet: the new `--installed-path {{skills_dir}}/<name>` contract can verify the wrong physical directory for valid skill names, and CI is still not green/actionable. 

### Response pass

* Addressed: The previous blocker about rejecting valid names with spaces is fixed. `validate_skill_name()` now uses a deny-list approach and accepts names such as `"Convex Best Practices"` while rejecting empty values, path separators, wildcards, control chars, leading hyphens, and reserved `.` / `..`. 
* Addressed: The weak missing-`npx` success path is fixed. Missing `npx` is now a hard error in non-dry-run mode, and lock verification is deterministic by re-reading `skills-lock.json` instead of parsing `npx skills list` output. 
* Addressed: Focused helper coverage was added and wired into `package.json` and CI. The new suite covers the #117 lock-patch regression, names with spaces, bad-name rejection, missing `npx`, `cwd=root`, and `--installed-path` behavior.  
* Addressed: The prior short-path invocation in `/aif` is no longer present; the current Skill Acquisition Strategy uses the full helper path. 
* Addressed: `npx skills remove` now runs with `cwd=str(root)`, so `--root` and the upstream CLI operate in the same project context. 
* Still unresolved: CI is still `completed/action_required`, and no jobs are visible through the API. This is still a merge gate for a security cleanup PR.
* New blocker found: the new `--installed-path {{skills_dir}}/<name>` instruction can point at a non-canonical path for valid skill names, so the helper can report successful physical cleanup while the real blocked skill directory remains.

### What looks good

* The helper is much safer than the initial version: it no longer silently succeeds when `npx` is unavailable, no longer depends on ANSI-formatted `skills list` output, and patches the lock deterministically. 
* The new tests are targeted and cheap: they use a fake `npx`, cover the previous review items, and are now part of the CI workflow.  
* Base freshness is fixed. The current head is `ahead_by: 5` and `behind_by: 0` against `2.x`; the previous stale-base concern no longer applies.

### Must fix

1. [P1] `--installed-path {{skills_dir}}/<name>` can verify the wrong directory — `skills/aif/SKILL.md`, `skills/aif-skill-generator/SKILL.md`, `SECURITY-SCANNING.md`

   The new `--installed-path` support is useful, but the prompt contract now constructs the path as `{{skills_dir}}/<name>`. That is not always the path the upstream `skills` CLI installs. Upstream installs sanitize the directory name: disk installs use `sanitizeName(skill.name)`, and well-known/blob installs use `sanitizeName(skill.installName)`. For example, `"Convex Best Practices"` maps to `convex-best-practices`, not `Convex Best Practices`.     

   Broken scenario:

   * A blocked skill has logical lock key / CLI name `"Convex Best Practices"`.
   * Its real installed directory is `.claude/skills/convex-best-practices`.
   * `/aif` runs:

     ```bash
     cleanup-blocked-skill.py --skill "Convex Best Practices" --installed-path .claude/skills/Convex Best Practices
     ```
   * If `npx skills remove` fails or partially fails, the helper checks the unsanitized path, sees that `.claude/skills/Convex Best Practices` is absent, downgrades `cli_failed` to success, and exits `0`.
   * The real blocked directory `.claude/skills/convex-best-practices` may still exist.

   This matters because `--installed-path` is now used to make the helper’s exit code “exact.” With the current prompt contract, that exactness is false for part of the valid skill-name surface that this PR just re-enabled.

   Expected fix:

   * In all skill docs, pass the actual scan target path to the helper:

     ```bash
     --installed-path <installed-path>
     ```

     or otherwise explicitly require the agent to reuse the same `<installed-skill-path>` that was passed to `security-scan.py`.
   * Do not synthesize the physical path from `<name>` unless the helper also implements the upstream `sanitizeName()` logic and clearly treats that as the canonical installed directory.
   * Add a regression test for a spaced/display name with `--installed-path`, where the real leftover directory is the sanitized path. The helper must exit `1` if the real sanitized directory still exists.
   * Add a prompt-contract regression check or grep-style test that the skill docs use `<installed-path>` / `<installed-skill-path>` for `--installed-path`, not `{{skills_dir}}/<name>`.

2. [P1] Current head still does not have a green CI merge signal — GitHub Actions

   The branch is no longer behind base, but the current workflow run for `e507c7d` is still `completed/action_required`, and the jobs list is empty through the API. That means I cannot verify build, lint, existing test suites, or the new cleanup-helper suite from CI.

   Expected fix:

   * Approve/rerun the workflow if this is GitHub’s first-time-contributor `action_required` gate.
   * Make the required CI run green on the current head.
   * If a workflow failure or non-run is unrelated, link the maintainer-approved run or base-branch evidence that proves it is unrelated.

### Should fix

1. [P2] PR body is now stale

   The PR body still says the helper “Verifies via `npx skills list`,” but the implementation now verifies by re-reading the patched lock file, and the behavior now also includes optional `--installed-path` physical-directory verification. The code direction is better than the body says, but the PR description should match the final behavior.  

   Expected fix:

   * Update the PR body to say verification is lock-file based.
   * Add a short note describing `--installed-path` and its intended caller contract.
   * Update the test plan checkboxes once `npm test` / CI actually pass.

### Nits

None.

### Context gates

* Architecture: pass — still isolated to skill prompt contracts, one helper, and focused tests; no broader runtime/package architecture change.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths.
* Docs / skill contract sync: error — the new `--installed-path` docs synthesize a path from `<name>` instead of reusing the actual installed path scanned earlier.  
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: error — the lock cleanup is stronger, but physical cleanup verification can false-pass for valid names whose installed directory is sanitized.
* Feature flag / risk gating: n/a — no runtime feature flag required for this helper/prompt security fix.
* Env / Turbo / process plumbing: n/a — no env/Turbo/process boundary changes.
* Managed assets / ownership: warn — the helper intentionally mutates `skills-lock.json`; the mutation path looks bounded, but the physical-directory verification contract needs correction.
* Migration / backward compatibility: warn — valid names with spaces are now accepted, but the new installed-path contract still mishandles their filesystem representation.
* Dependency / lockfile / production install: pass — no dependency or package lock change.
* PR size / decomposition: pass — 7 files, but the added test file is directly tied to the security fix.
* Stack / merge order / base freshness: pass — current head is not behind `2.x`.
* CI / audit: error — CI is still `action_required`; no green job evidence is visible.

### Verification

* Checked: current PR metadata/head, changed files, helper implementation, `/aif` call sites, skill-generator/security-scanning call sites, package script wiring, CI workflow wiring, PR comments, previous review items, upstream `skills` install path sanitization, and the supplied review rubric.  
* Ran locally: not run. I reviewed the source and test logic through GitHub context only.
* CI: current known state is `CI completed/action_required` for `e507c7d`; no jobs are visible from the API.
* Not verified: full `npm test`, the new `npm run test:cleanup-helper`, live `npx skills` end-to-end behavior, and maintainer-only required checks.

### Verdict

Verdict: REQUEST_CHANGES until the `--installed-path` contract uses the actual installed/scanned path or derives the canonical sanitized path safely, and CI is rerun/green on the current head.

Meacue:
## Update: addressed in `2e4ecad`

@ichinya addressed the `--installed-path {{skills_dir}}/<name>` blocker from your last review. Could you re-verify?
PR body updated to reflect the actual current behavior (lock-file verification, optional `--installed-path` physical check, the sanitization caveat in the caller contract).

@lee-to there's a **"1 workflow awaiting approval"** banner directly in this PR with an **Approve and run** button (GitHub's first-time-contributor gate on fork PRs). Could you click it when you have a moment?

ichinya:
## Code review

Follow-up review for PR #119 at current head 2e4ecad368a930980e0c86316ac4cbf6a8c03c6d. The PR now replaces the old BLOCKED-skill `rm -rf` path with a cleanup helper that patches `skills-lock.json`, verifies the lock state, optionally verifies the installed directory, updates the affected skill docs, and adds a focused helper regression suite. This directly targets #117, where deleting only the physical directory left a stale lock entry that could resurrect a security-blocked skill through `experimental_install`.  

The previous blockers are mostly addressed now. I still found one runtime blocker before merge: for valid skills whose logical lock/display name differs from the sanitized on-disk directory name, the helper can now detect that cleanup failed, but it still does not actually remove the blocked directory.

### Response pass

* Addressed: The helper now accepts valid skill names with spaces while rejecting genuinely unsafe values such as path separators, wildcards, control characters, leading hyphens, and reserved `.` / `..`. 
* Addressed: Missing `npx` is now a hard failure outside `--dry-run`, and lock verification is deterministic by re-reading `skills-lock.json` instead of parsing `npx skills list` output. 
* Addressed: The short unresolved helper invocation in `/aif` is fixed; current call sites use the full helper path. 
* Addressed: `npx skills remove` is now run with `cwd=<root>`, so `--root` and the upstream CLI operate in the same project context. 
* Addressed: The `--installed-path` prompt contract now tells agents to pass the same path used for `security-scan.py`, not `{{skills_dir}}/<name>`, and a grep regression test covers that contract.      
* Addressed: Base freshness and CI are fixed. The current head is `behind_by: 0`, and the current CI run completed successfully with build, lint, `npm test`, init/update smoke tests, and cleanup-helper tests all green.
* New blocker found: The helper still calls upstream `npx skills remove` with the logical lock/display name. Upstream removal scans installed directory names and matches them case-insensitively, but not through the install-time sanitizer. That means a skill such as `"Convex Best Practices"` installed at `convex-best-practices` will not be removed by `npx skills remove -s "Convex Best Practices"`; the helper will patch the lock and then return `1` because `--installed-path` still exists, leaving the blocked skill physically present.   

### What looks good

* The lock-file fix is now bounded and deterministic: it patches only `skills[<name>]`, atomically writes the lock file, then re-reads the lock to verify the key is absent. 
* The docs/prompt contract now correctly tells agents to reuse the actual scanned path for `--installed-path`, which fixes the previous false-pass risk for names whose disk path is sanitized.  
* The new helper tests are focused on the real failure modes: stale lock removal, spaced names, invalid names, missing `npx`, `cwd` propagation, `--installed-path`, sanitized leftover behavior, and prompt-contract regression. 
* The broad `Bash(rm -rf *)` permission is still removed from the affected skill frontmatter and replaced with a narrower `Bash(python *cleanup-blocked-skill*)` allowance. 

### Must fix

1. [P1] Helper verifies but does not remove sanitized-path leftovers for logical names with spaces — `skills/aif-skill-generator/scripts/cleanup-blocked-skill.py`

   The current helper calls:

   ```python
   [npx, 'skills', 'remove', '-s', skill, '-y']
   ```

   where `skill` is the logical lock key / skill name. For GitHub/local disk installs, upstream install sanitizes the skill name before choosing the directory name, so `"Convex Best Practices"` is installed under `convex-best-practices`. Upstream removal, however, scans installed directory names and matches them only by lowercased equality against the provided argument; it does not apply `sanitizeName()` to the argument before matching. If no match is found, `removeCommand()` returns, and the CLI wrapper exits `0` through `main().finally(... process.exit(0))`.      

   Broken scenario:

   * External skill frontmatter / lock key is `"Convex Best Practices"`.
   * Actual scanned path is `.claude/skills/convex-best-practices`.
   * Security scan returns BLOCKED.
   * The skill prompt now correctly calls:

     ```bash
     cleanup-blocked-skill.py --skill "Convex Best Practices" --installed-path .claude/skills/convex-best-practices
     ```
   * Upstream `npx skills remove` does not match `Convex Best Practices` to `convex-best-practices`, returns effectively successful/no-op, and does not remove the directory.
   * The helper patches `skills-lock.json`, sees `.claude/skills/convex-best-practices` still exists, and exits `1`.

   Exiting `1` is better than a false success, but it still leaves the security-blocked skill physically installed. Since this PR removed the broad `rm -rf` allowance from the agent prompt, the advertised cleanup path no longer has a working fallback for this valid name surface. That means the stale-lock resurrection issue is fixed, but the immediate blocked-skill-on-disk issue remains for sanitized-name cases.

   Expected fix:

   * Keep patching the lock by the logical `--skill` key.
   * Also remove the physical directory represented by `--installed-path`, either by:

     * calling `npx skills remove` with the installed directory basename as a second attempt when the logical-name attempt leaves `--installed-path` present; or
     * having the Python helper remove `--installed-path` directly after strict safety validation.
   * If deleting directly, validate that the target is an absolute or `--root`-relative directory inside a known project/global skills directory, reject path traversal/symlink escape if applicable, and require a plausible skill directory marker such as `SKILL.md` before deletion.
   * Add a regression test where:

     * lock key is `"Convex Best Practices"`;
     * actual installed path is `.claude/skills/convex-best-practices`;
     * fake `npx` simulates upstream no-match with exit `0` and no deletion;
     * helper exits `0`, lock key is gone, and `.claude/skills/convex-best-practices` is gone.
   * Add a negative safety test that `--installed-path ../outside` or an unrelated directory is rejected and not deleted.

### Should fix

1. [P2] Update PR body/test-plan status after the current CI run

   The PR body still says CI is `action_required` and leaves `npm test` / CI unchecked, but the current head has a successful CI run and the test job shows build, lint, `npm test`, smoke tests, and cleanup-helper tests all succeeded. This is not a code blocker, but the PR body should reflect the actual merge evidence.

   Expected fix:

   * Check off CI and `npm test` in the PR body.
   * Remove or update the stale `action_required` note.
   * Keep the note about pre-existing Windows-only local failures only if it is still useful to reviewers.

### Nits

None.

### Context gates

* Architecture: pass — the change is still isolated to skill prompt contracts, one helper script, and a focused test script.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths.
* Docs / skill contract sync: pass — the prompt/docs now consistently tell agents to reuse the actual scanned installed path for `--installed-path`.     
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: error — the helper can still leave a security-blocked skill physically installed when the logical lock key does not match the sanitized installed directory name.
* Feature flag / risk gating: n/a — no runtime feature flag is needed for this helper/prompt security fix.
* Env / Turbo / process plumbing: n/a — no env/Turbo/process-boundary change.
* Managed assets / ownership: warn — lock mutation is bounded, but physical cleanup still depends on upstream remove matching the same key surface as the lock.
* Migration / backward compatibility: warn — names with spaces are accepted and lock cleanup works, but physical cleanup still fails for the same class unless the helper handles sanitized directory basenames.
* Dependency / lockfile / production install: pass — no dependency or lockfile change.
* PR size / decomposition: pass — 7 files, with most added lines in targeted regression tests.
* Stack / merge order / base freshness: pass — current head is `behind_by: 0` against `2.x`.
* CI / audit: pass — current GitHub Actions CI run for the head is successful.

### Verification

* Checked: current PR head/body/diff, previous review comments, linked issue #117, helper implementation, `/aif` call sites, `/aif-skill-generator` call sites, security-scanning reference, cleanup-helper tests, CI workflow, package script wiring, base freshness, current CI run/job steps, and upstream `vercel-labs/skills` install/remove behavior.    
* Ran locally: not run; source and CI evidence were reviewed through GitHub context only.
* CI: current known state is green for the head; the `test` job completed successfully through the cleanup-helper test step.
* Not verified: live end-to-end `npx skills` behavior against a real external skill with a spaced logical name; the blocker above is inferred from upstream source behavior and should be covered by a fake-`npx` regression.

### Verdict

Verdict: REQUEST_CHANGES until the helper actually removes the `--installed-path` directory for sanitized-name cases, or safely retries removal using the installed directory basename, and the regression test covers logical-name/physical-name mismatch.

Meacue:
## Update: addressed in `a4cb2cd`

**P1**: Helper itself now removes the directory via `safe_remove_installed` (Python `rmtree` with strict safety validation). `npx skills remove` kept as best-effort first pass. New tests N1–N8 cover each rejection path; flipped Test 11(b) is the sanitized-leftover end-to-end case you asked for.

**P2**: PR body refreshed - CI / `npm test` checked, stale `action_required` note removed.

@ichinya could you please re-verify?

ichinya:
## Code review

Follow-up review for PR #119 at current head a617c956674ab766835f512a123c5c7529a3e89d. The PR now replaces the old BLOCKED-skill `rm -rf` path with a dedicated cleanup helper that clears `skills-lock.json`, verifies lock state, and physically removes the scanned installed skill directory via `--installed-path`. The prior blocker about sanitized names is largely addressed by making the helper itself the physical-removal guarantor instead of relying on upstream `npx skills remove`. Merge is close, but I still found one deletion-boundary blocker before approval. 

### Response pass

* Addressed: The previous valid-name-surface blocker is fixed. The helper now accepts names with spaces and rejects unsafe values such as path separators, wildcards, control characters, leading hyphens, and reserved `.` / `..`. 
* Addressed: Missing `npx` is now a hard failure outside `--dry-run`, and `npx skills remove` runs with `cwd=<root>`. 
* Addressed: Lock verification is now deterministic by re-reading `skills-lock.json`, not by parsing `npx skills list` output. 
* Addressed: The `--installed-path` prompt contract now consistently tells agents to reuse the same path passed to `security-scan.py`, not synthesize `{{skills_dir}}/<name>`.     
* Addressed: The prior sanitized-directory blocker is addressed in behavior: `safe_remove_installed()` now physically removes the supplied `--installed-path`, so `Convex Best Practices` can be removed from `convex-best-practices` when the scanned path is passed. 
* Addressed: CI is now green on the current head, including build, lint, `npm test`, init/update smoke tests, and cleanup-helper tests.
* New blocker found: `safe_remove_installed()` still accepts too broad a deletion boundary: any directory under repo root with a `skills` path segment and `SKILL.md` marker can be deleted, even if it is not an installed agent skill and even if the lock entry was absent.

### What looks good

* The helper is now the actual cleanup authority for physical deletion. That avoids the upstream mismatch where `npx skills remove -s "Convex Best Practices"` does not match a sanitized directory basename like `convex-best-practices`. 
* The deletion flow has meaningful defense-in-depth checks: symlink / Windows reparse-point rejection, root containment, root equality guard, `SKILL.md` marker requirement, and read-only cleanup retry.  
* The test suite now covers the earlier review cases and the round-4 physical-removal scenario, including sanitized leftover removal and prompt-contract grep coverage. 
* Current CI is green for the head under review, with the cleanup-helper test step included and passing.

### Must fix

1. [P1] `safe_remove_installed()` can delete user/source directories that merely contain a `skills` segment — `skills/aif-skill-generator/scripts/cleanup-blocked-skill.py`

   The new physical deletion path is directionally correct, but the boundary is still too broad for a helper that replaces broad `rm -rf` permission. `safe_remove_installed()` accepts any resolved directory under `--root` if the relative path contains a `skills` segment and the directory has a `SKILL.md` marker. It does not require the path to be an actual installed agent skill path like `.<agent>/skills/<skill>`, and it also runs even when `patch_lock()` reports that the requested skill key was absent from `skills-lock.json`.  

   Broken scenario:

   * A project or this repository itself contains source/docs skills such as `skills/aif/SKILL.md`, `skills/aif-skill-generator/SKILL.md`, or `docs/skills/example/SKILL.md`.
   * The helper is invoked with a typo or wrong scan target, for example:

     ```bash
     cleanup-blocked-skill.py --skill typo --root . --installed-path skills/aif
     ```
   * `patch_lock()` returns a non-mutating “skill not present” result instead of failing.
   * `verify_lock_clean()` returns true because the key is absent.
   * `safe_remove_installed()` sees the path is inside root, contains a `skills` segment, has `SKILL.md`, and deletes the source/user-owned directory.

   This matters because the PR intentionally removes broad shell deletion permission from the skill prompts. The replacement helper should be narrower than `rm -rf`, but the current path policy still permits deletion of repository-owned or user-owned skill-like directories that are not installed agent skills.

   Expected fix:

   * Tighten the installed-path boundary. At minimum, require the relative shape to be an installed agent-skill directory, not just “contains `skills`”:

     * `.<agent-config>/skills/<skill-dir>` for project installs; or
     * an explicit, documented global-skill root if global cleanup is intentionally supported.
   * Reject root-level source paths such as `skills/<name>` and nested docs paths such as `docs/skills/<name>`.
   * Couple the deletion target to the requested skill identity, for example by checking one of:

     * basename equals `sanitizeName(--skill)`;
     * `SKILL.md` frontmatter `name` matches `--skill`;
     * the lock entry existed and the `--installed-path` basename matches the installed key expected by upstream.
   * Decide explicitly what should happen when the lock entry is absent. If deletion is still allowed for idempotency, it must pass the stricter installed-agent-path and identity checks.
   * Add regression tests:

     * `skills/source-skill/SKILL.md` under repo root is rejected and preserved.
     * `docs/skills/source-skill/SKILL.md` is rejected and preserved.
     * `.claude/skills/foo/SKILL.md` is accepted and removed.
     * A mismatched `--skill foo --installed-path .claude/skills/bar` is rejected unless there is a deliberate documented reason to allow it.

### Should fix

1. [P2] Branch is currently behind `2.x` by two commits

   The behind-base diff appears unrelated to this PR (`skills/aif-improve/tests/sequential-plan-discovery.spec.yaml` only), so I would not block on it as a code issue. If branch protection or merge queue requires an up-to-date base, rebase/update and rerun CI before merge.

### Nits

None.

### Context gates

* Architecture: pass — the change remains isolated to the helper, skill prompt contracts, and focused tests.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths.
* Docs / skill contract sync: pass — the docs now consistently instruct agents to pass the same scanned installed path to `--installed-path`.
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: error — physical deletion now works, but the deletion boundary is too broad for user/source directories containing `skills/` and `SKILL.md`.
* Feature flag / risk gating: n/a — no runtime feature flag is needed for this helper/prompt security fix.
* Env / Turbo / process plumbing: n/a — no env/Turbo/process-boundary change.
* Managed assets / ownership: error — the helper can delete repository/user-owned skill-like directories that are not managed installed agent skills.
* Migration / backward compatibility: pass — valid spaced skill names are accepted, and sanitized physical paths can now be removed when the scanned path is passed.
* Dependency / lockfile / production install: pass — no dependency or lockfile change.
* PR size / decomposition: pass — the PR is larger now, but the bulk is directly targeted regression coverage.
* Stack / merge order / base freshness: warn — current head is behind `2.x` by two commits, apparently unrelated; update if branch protection requires it.
* CI / audit: pass — current CI for the head is green, including cleanup-helper tests.

### Verification

* Checked: current PR head/body/diff, previous review comments, linked issue #117, helper implementation, `/aif` and `/aif-skill-generator` call sites, security-scanning reference, cleanup-helper tests, package script wiring, CI workflow, current CI run/job steps, base freshness, and the supplied review rubric.   
* Ran locally: not run; source and CI evidence were reviewed through GitHub context only.
* CI: current known state is green for `a617c956674ab766835f512a123c5c7529a3e89d`; the `test` job completed successfully through build, lint, validation, smoke tests, and cleanup-helper tests.
* Not verified: live end-to-end cleanup against a real external skill; the blocker above is inferred from the helper’s own path checks.

### Verdict

Verdict: REQUEST_CHANGES until `safe_remove_installed()` is restricted to actual installed agent-skill directories and the regression tests prove source/user-owned `skills/*` paths cannot be deleted.

Meacue:
## Update: addressed in `cd8e37a`, `f48e27a`

**P1**: Path boundary tightened via new `validate_installed_skill_path()`:
- direct child of a known agent skillsDir (built-in 15-entry list + `.ai-factory.json` `agents[].skillsDir` for extensions)
- identity coupling: sanitized `basename.lower()` OR SKILL.md frontmatter `name:` must match `--skill`
- lock-absent path still required to pass both shape and identity checks

New tests N9-N15 cover the 4 cases you listed + unregistered `.cache/skills/foo` + `.ai-factory.json` extension + dry-run validation.

**P2**: Merged `upstream/2.x` in `f48e27a`.

@ichinya I've lost track of how many times I've tried to merge, maybe I'll get lucky this time.

ichinya:
## Code review

Follow-up review for PR #119 at current head f48e27a62706409246b112c5b48db5246041110f. The PR now replaces the old BLOCKED-skill `rm -rf` path with a dedicated cleanup helper that clears `skills-lock.json`, validates the installed path, and physically removes the blocked skill directory. The previous blockers are mostly addressed, including the source/user-owned `skills/*` deletion boundary. I still found one runtime blocker in the symlinked install path, and the current head does not yet have a green CI run. 

### Response pass

* Addressed: The helper now accepts the upstream skill-name surface, including names with spaces, while rejecting unsafe names such as path separators, wildcards, control characters, leading hyphens, and reserved `.` / `..`. 
* Addressed: The helper now validates `--installed-path` as a direct child of known project agent `skillsDir` roots, including extension-defined roots from `.ai-factory.json`, instead of accepting any path that merely contains a `skills` segment. It also checks basename/frontmatter identity against `--skill`. 
* Addressed: The helper is now the physical cleanup authority when `--installed-path` is provided, so sanitized names like `"Convex Best Practices"` can be removed from `.claude/skills/convex-best-practices` even when upstream `npx skills remove` is a no-op. 
* Addressed: The test suite now includes safety coverage for non-agent paths, path traversal, missing `SKILL.md`, symlink/junction rejection, identity mismatch, sanitized-name cleanup, and prompt-contract regressions.  
* Still unresolved: Current head CI is `action_required`, with no jobs visible through the API; it is not a green merge signal for this head.
* New blocker found: The helper rejects symlink/junction installed paths before running `npx skills remove`, but upstream `skills` can legitimately install project skills by copying to a canonical directory and symlinking agent-specific directories. That makes a valid scanned path fail closed before cleanup.  

### What looks good

* The helper now validates `--installed-path` before touching the lock file, so invalid deletion targets do not get a stale-lock cleanup first.
* The physical deletion path is much narrower than `rm -rf`: it requires project-root containment, a known direct agent skills root, identity match, directory target, and a `SKILL.md` marker before deletion. 
* The PR body now clearly documents why the helper owns physical removal and why callers must pass the same path that was scanned rather than synthesizing from the logical skill name. 
* The broad `Bash(rm -rf *)` allowance is still removed from the affected skills and replaced with the narrower helper invocation. 

### Must fix

1. [P1] Valid symlinked installed paths are rejected before cleanup — `skills/aif-skill-generator/scripts/cleanup-blocked-skill.py`

   The helper currently rejects any symlink or Windows junction in `_existing_path_prefixes()` before it calls `npx skills remove` or patches the lock. That is safe against malicious redirects, but it also blocks a normal upstream installation shape: in symlink mode, upstream `skills` writes the skill into the canonical `.agents/skills/<name>` directory and creates agent-specific symlinks such as `.claude/skills/<name>`.  

   Broken scenario:

   * A project has a symlinked install: `.claude/skills/foo -> ../../.agents/skills/foo`.
   * `/aif` scans the actual path it sees, `.claude/skills/foo`, and the scan returns BLOCKED.
   * The prompt correctly calls:

     ```bash
     cleanup-blocked-skill.py --skill foo --installed-path .claude/skills/foo
     ```
   * `validate_installed_skill_path()` rejects the symlink prefix before `npx skills remove` gets a chance to remove the canonical directory and links.
   * Result: the helper exits `1`, the lock entry remains, and the blocked skill remains installed.

   This matters because the helper is now the only allowed cleanup path after dropping `Bash(rm -rf *)`. A valid upstream install mode should not leave the agent unable to clean up a blocked skill.

   Expected fix:

   * Keep rejecting symlinks/junctions that resolve outside the project or outside known project agent skills roots.
   * Allow the normal managed symlink case when the symlink target resolves to a direct child of an allowed project agent skills root and the target passes the same identity checks.
   * Either:

     * validate the symlink target and delete the canonical target plus the link; or
     * run the upstream `npx skills remove` first for this managed-symlink case, then verify both the symlink path and resolved canonical target are gone.
   * Add regression tests:

     * `.agents/skills/foo/SKILL.md` plus `.claude/skills/foo -> ../../.agents/skills/foo` is accepted and cleaned.
     * A symlink/junction to `../outside/foo` is still rejected and preserved.
     * The lock entry is cleared only after the managed symlink/canonical cleanup path succeeds.

### Should fix

1. [P2] Avoid hard-coding built-in agent skills roots long-term

   `_BUILTIN_AGENT_SKILLS_DIRS` currently mirrors the built-in agent registry, and the list appears complete for the current built-ins. However, this is a second source of truth: future agent additions or skillsDir changes can make the cleanup helper reject valid installed paths until someone remembers to update this Python list. The `.ai-factory.json` extension hook helps for runtime extension agents, but built-ins still drift manually.  

   Expected fix:

   * Track as a follow-up unless you want to address it here.
   * Prefer generating this allowlist from the same registry used by `src/core/agents.ts`, or add a focused test that compares the helper’s built-in list against the built-in `skillsDir` values.

### Nits

None.

### Context gates

* Architecture: pass — the change remains isolated to the helper, skill prompt contracts, CI wiring, and focused tests.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths.
* Docs / skill contract sync: pass — the prompts consistently instruct callers to reuse the same path that was passed to `security-scan.py`.
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: error — unmanaged symlink/junction rejection is good, but managed upstream symlink installs currently cannot be cleaned.
* Feature flag / risk gating: n/a — no runtime feature flag is needed for this helper/prompt security fix.
* Env / Turbo / process plumbing: n/a — no env/Turbo/process-boundary change.
* Managed assets / ownership: warn — the direct-child and identity checks fix the previous broad deletion boundary, but managed symlink ownership still needs handling.
* Migration / backward compatibility: warn — copy-mode and canonical-path installs are covered; symlink-mode installs remain a compatibility gap.
* Dependency / lockfile / production install: pass — no dependency or lockfile change.
* PR size / decomposition: warn — the PR is now larger, but the added size is primarily targeted regression coverage for a security cleanup helper.
* Stack / merge order / base freshness: pass — current head is `behind_by: 0` against `2.x`.
* CI / audit: error — current head CI is `action_required`, not green.

### Verification

* Checked: current PR head/body/diff, previous review comments, helper implementation, skill prompt call sites, current cleanup-helper tests, CI workflow status, base freshness, upstream `skills` symlink install behavior, and the supplied review rubric.   
* Ran locally: not run; source and CI evidence were reviewed through GitHub context only.
* CI: current known state for f48e27a62706409246b112c5b48db5246041110f is `completed/action_required`; no jobs are visible from the API.
* Not verified: live end-to-end cleanup against a real symlink-mode upstream `npx skills` install.

### Verdict

Verdict: REQUEST_CHANGES until managed symlinked installed paths can be cleaned safely and the current head has a green CI run.

ichinya:
Thanks for the continued iteration here — this is much closer now.

The previous issues around stale lock cleanup, names with spaces, using the actual scanned path, and avoiding broad `rm -rf` are now addressed in a much safer direction. I also like that the helper is becoming the single bounded cleanup path instead of relying on the upstream CLI’s imperfect matching behavior.

I still have one concern before I’d be comfortable merging: the current safety check rejects symlink/junction paths up front, but symlinked project installs are a normal `skills` install shape. In that case the scanned path may be something like `.claude/skills/foo`, while the real canonical copy is under `.agents/skills/foo`. If `/aif` scans the symlink path and then calls the helper with that same path, the helper will reject before it can clean either the symlink or the canonical directory.

I think the right fix is to keep rejecting unsafe symlinks that escape the project or point outside known agent skill roots, but allow the managed symlink case when the resolved target is also a valid installed skill directory. The helper can then remove both the link and the canonical target, or let `npx skills remove` handle the link cleanup and verify both paths are gone afterwards.

Suggested regression coverage:
- `.agents/skills/foo/SKILL.md` with `.claude/skills/foo -> ../../.agents/skills/foo` is accepted and cleaned;
- a symlink/junction to an outside directory is still rejected and preserved;
- the lock entry is cleared only after the managed symlink cleanup succeeds.

Meacue:
## Update: addressed the managed symlink blocker in `b079c26`.

Changes:
- `--installed-path` now accepts the normal upstream symlinked install shape, e.g. `.claude/skills/foo -> ../../.agents/skills/foo`, when both the symlink path and resolved canonical target stay inside the project and under known agent `skillsDir` roots.
- The helper removes the canonical installed directory and then removes the managed symlink/junction path without following it.
- Escaping symlinks/junctions still fail closed and preserve both the target and the lock entry.
- Lock cleanup still happens only after physical cleanup succeeds, so a failed symlink cleanup cannot clear `skills-lock.json` first.
- Added regression coverage for managed symlink/junction cleanup, outside symlink/junction rejection, and lock preservation when managed cleanup fails.

CI is green on `b079c26104e1057b0177206dc304a6d32fddfd6d`.

@ichinya could you please re-verify?

ichinya:
## Code review

Follow-up review for PR #119 at current head b079c26104e1057b0177206dc304a6d32fddfd6d. This PR replaces the previous BLOCKED-skill `rm -rf` cleanup path with a bounded helper that removes the installed skill directory, patches `skills-lock.json`, validates the lock state, and updates the affected skill prompt contracts. The earlier blockers around valid names with spaces, stale lock cleanup, unsafe `{{skills_dir}}/<name>` path synthesis, source-owned `skills/*` deletion, and managed symlink installs are addressed now. I do not see meaningful code blockers remaining. 

### Response pass

* Addressed: The helper now accepts the upstream skill-name surface, including names with spaces, while rejecting unsafe names such as path separators, wildcards, control characters, leading hyphens, and reserved `.` / `..`.
* Addressed: Cleanup no longer relies on upstream `npx skills remove` as the sole physical-removal mechanism. The helper owns physical cleanup when `--installed-path` is supplied, then patches and verifies the lock file. 
* Addressed: The broad deletion boundary from the previous round is tightened. `--installed-path` must resolve to a direct child of a known project agent skills directory, and identity is checked through sanitized basename or `SKILL.md` frontmatter. 
* Addressed: Managed symlink/junction installs are handled: the helper validates the resolved canonical target, removes it, and then removes the agent-specific reparse point/link. Escaping symlinks remain rejected. 
* Addressed: CI is green on the current head, including build, lint, skill validation, init/update smoke tests, and cleanup-helper tests.
* Still unresolved: No code blocker from my side.
* New blocker found: None.

### What looks good

* The helper is now a bounded replacement for `rm -rf`: it validates project-root containment, known agent skills roots, identity, `SKILL.md` presence, symlink/junction behavior, and physical cleanup result before considering cleanup successful.
* The operation order is safer now: with `--installed-path`, physical cleanup happens before lock patching, so a failed delete no longer clears the lock while leaving the blocked skill installed.
* The prompt contract is consistent: callers are told to pass the exact path previously scanned by `security-scan.py`, not a path synthesized from the logical skill name.
* The regression coverage is substantial and targeted: spaced names, stale lock removal, missing `npx`, sanitized-name cleanup, managed symlink cleanup, escaping symlinks, source-path rejection, identity mismatch, extension skillsDir support, and prompt-contract drift are covered.

### Must fix

I do not have meaningful code blockers anymore.

### Should fix

1. [P2] Keep the PR body test-count / safety-summary wording in sync

   The PR body still describes “20 focused regression tests” and “Safety checks N1–N8”, but the test file has grown beyond that with additional managed-symlink, source-path, extension-root, dry-run, and drift checks. This is not a merge blocker, but the body should not under-describe the current safety coverage.

   Expected fix:

   * Either update the count/numbered ranges to match the current test file, or make the wording count-free, for example “focused regression tests covering…”.
   * Keep the high-level explanation of `safe_remove_installed`; that context is useful for future reviewers.

### Nits

None.

### Context gates

* Architecture: pass — the change is still isolated to the cleanup helper, prompt contracts, CI wiring, and focused tests.
* Package CHECKLIST: n/a — I did not find a package-specific CHECKLIST for these touched paths.
* Docs / skill contract sync: pass — the affected skill docs now consistently route BLOCKED cleanup through the helper and tell callers to reuse the scanned installed path.
* API/web/client sync: n/a — no API, web, or generated client contract change.
* Security surfaces: pass — the previous broad deletion and symlink concerns are addressed with direct-agent-root, identity, marker, containment, and reparse-point checks.
* Feature flag / risk gating: n/a — no runtime feature flag is needed for this helper/prompt security fix.
* Env / Turbo / process plumbing: n/a — no env/Turbo/process-boundary change.
* Managed assets / ownership: pass — source-owned `skills/*` and docs-owned `docs/skills/*` paths are rejected; cleanup is limited to known project agent skills roots.
* Migration / backward compatibility: pass — valid spaced names and sanitized installed paths are supported, and missing installed paths remain idempotent.
* Dependency / lockfile / production install: pass — no dependency or lockfile change.
* PR size / decomposition: warn — the diff is large now, but the bulk is targeted regression coverage for a security cleanup helper.
* Stack / merge order / base freshness: pass — current head is not behind `2.x`.
* CI / audit: pass — current CI for the head is green.

### Verification

* Checked: current PR head/body/diff, helper implementation, affected `SKILL.md` call sites, cleanup-helper tests, prior review threads/comments, base freshness, and current CI status. 
* Ran locally: not run; reviewed through GitHub context and CI evidence.
* CI: current known state is green for b079c26104e1057b0177206dc304a6d32fddfd6d; the `test` job completed successfully through build, lint, validation, smoke tests, and cleanup-helper tests.
* Not verified: live end-to-end cleanup against a real external `skills` install; the fake-`npx` regression suite covers the important helper behavior.

### Verdict

Verdict: APPROVE / no meaningful blockers.

lee-to:
## Maintainer review

I re-checked current head `b079c26104e1057b0177206dc304a6d32fddfd6d` after the latest approval, focusing on the cleanup helper, prompt call sites, symlink/junction handling, lock-patch ordering, and regression coverage.

No blocking issues found. The current implementation addresses the stale-lock resurrection bug from #117 and the previous review concerns:

- BLOCKED cleanup no longer relies on broad `rm -rf` or on upstream `npx skills remove` matching as the only physical-removal mechanism.
- `--installed-path` is validated before deletion, and physical cleanup completes before the lock entry is patched.
- Valid names with spaces and sanitized on-disk paths are covered.
- Managed symlink/junction installs are accepted only when the link and canonical target stay under known project agent `skillsDir` roots; escaping redirects still fail closed.
- Prompt docs consistently tell agents to pass the exact path that was scanned, not a synthesized `{{skills_dir}}/<name>` path.
- CI is green on the current head. I also ran `python3 -m py_compile` on the helper locally.

Non-blocking hardening note: consider rejecting or normalizing leading/trailing whitespace in `--skill` in a follow-up. Internal spaces are valid, but edge whitespace could make the exact lock-key patch miss while the sanitized installed path still validates. This is not a merge blocker for the current prompt flow, which passes the CLI skill name directly.

Verdict: APPROVE / ready to merge.

PR #122, opened 2026-05-25
fix(skills): honor language policy in workflow skills

### LINKED ISSUE #121: aif-explore ignores configured language.ui because it does not load config.yaml or shared language
## Problem

Project `.ai-factory/config.yaml` can define:

```yaml
language:
  ui: russian
  artifacts: russian
  technical_terms: keep
```

But `$aif-explore` may still answer in English.

## Reproduction

1. In a project, set `.ai-factory/config.yaml` with `language.ui: russian`.
2. Run `$aif-explore` on a YouGile/task/code exploration request.
3. Observe the user-facing exploration summary.

## Expected

All user-facing responses from AIF skills should use `language.ui` from `.ai-factory/config.yaml`.

## Actual

`$aif-explore` produced an English response even though the project config requested Russian.

## Likely Cause

`aif-explore/SKILL.md` tells the agent to read project context files like `DESCRIPTION.md`, `ARCHITECTURE.md`, `RULES.md`, `RESEARCH.md`, `PLAN.md`, and `ROADMAP.md`, but it does not require reading `.ai-factory/config.yaml` or applying the shared language policy.

The shared policy exists and says to use `language.ui` for user-facing responses, but `aif-explore` does not reference it.

## Suggested Fix

Update `aif-explore` and other AIF skills to explicitly:

- read `.ai-factory/config.yaml` at startup when present;
- apply `language.ui` for all user-facing responses;
- apply `language.artifacts` for generated artifacts;
- keep commands, paths, identifiers, config keys, and code terms unchanged when `language.technical_terms: keep`.

Ideally this should be a shared mandatory prelude for all AIF skills, not duplicated per skill.

ichinya:
## Summary
- Fix `/aif-explore` so user-facing output follows `language.ui`, persisted research follows `language.artifacts`, and stable technical tokens remain preserved.
- Broaden the same language-policy contract across workflow artifact skills: plan, improve, fix, rules, reference, security checklist, and verify.
- Update docs/config matrices and add regression checks for the expanded language-policy coverage.

## Linked issue
Fixes #121

## Why
Issue #121 reported that `/aif-explore` ignored the configured UI language. During follow-up review, several adjacent workflow skills had the same incomplete language contract for reports or persisted artifacts.

## Validation
- `npm run build` passed
- `npm run lint` passed
- `git diff --check` passed
- `bash -n scripts/test-skills.sh` passed
- `npm test` ran: 116/122 passed; remaining failures are existing local environment issues (`python3` missing for aif-distillation helper tests and Windows npm resolver smoke failure)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -142,7 +142,7 @@ Current config keys in active use:
 
 - `paths.*` - artifact discovery for description, architecture, roadmap, research, RULES.md, plan files, fix plans, QA artifacts,
   references, security state, patches, evolutions, loop state, and rules
-- `language.ui` / `language.artifacts` - prompt language vs generated artifact language
+- `language.ui` / `language.artifacts` / `language.technical_terms` - prompt language, generated artifact language, and terminology handling
 - `git.enabled` / `git.base_branch` / `git.create_branches` / `git.branch_prefix` / `git.skip_push_after_commit` - planning, verification, and commit push behavior
 - `workflow.verify_mode` - default verification strictness
 - `rules.base` plus named `rules.<area>` entries - rules hierarchy
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -142,7 +142,7 @@ Current config keys in active use:
 
 - `paths.*` - artifact discovery for description, architecture, roadmap, research, RULES.md, plan files, fix plans, QA artifacts,
   references, security state, patches, evolutions, loop state, and rules
-- `language.ui` / `language.artifacts` / `language.technical_terms` - prompt language, generated artifact language, and terminology handling
+- `language.ui` / `language.artifacts` / `language.technical_terms` - prompt/report language, generated artifact language, and terminology handling while preserving commands, paths, identifiers, config keys, and raw errors where required
 - `git.enabled` / `git.base_branch` / `git.create_branches` / `git.branch_prefix` / `git.skip_push_after_commit` - planning, verification, and commit push behavior
 - `workflow.verify_mode` - default verification strictness
 - `rules.base` plus named `rules.<area>` entries - rules hierarchy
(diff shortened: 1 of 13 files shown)


lee-to (commented):
Review notes:

1. Medium: `language.ui` is declared, but several user-visible templates still remain literal English. For example, `skills/aif-plan/SKILL.md` says summaries and next-step guidance must use `ui_language`, but the next-step templates still say `Plan created with [N] tasks.` / `To start implementation, run:`. The same pattern appears in `/aif-improve` (`## Plan Refined`, `Changes applied`, `Ready to implement`) and `/aif-fix` (`## Fix Applied`, `Issue`, `Cause`, `Fix`). This can still reproduce the original class of bug for those skills. These templates should be marked as structure-only and rendered in `ui_language`, the same way `/aif-verify` now handles its report template.

2. Low/Medium: the `/aif-rules` config ownership contract is now contradictory. `docs/config-reference.md` correctly lists `/aif-rules` under Config Writers with `Yes, limited`, but the new reader row says `Writes config: No`. `scripts/test-skills.sh` also asserts the `Yes | No` reader row, so fixing the docs would break the regression test. Either remove `/aif-rules` from the reader-only table or make the test assert the limited writer contract instead.

ichinya:
Addressed the review notes in commit 08d754f:

- Marked the remaining `/aif-plan`, `/aif-improve`, and `/aif-fix` user-facing templates as structure-only and required their human-readable output to render in `ui_language`.
- Fixed the `/aif-rules` config matrix entry to `Writes config: Yes, limited` and updated the regression check to reject the contradictory `Yes | No` row.

Verification:
- `npm run build` passes.
- `git diff --check` passes.
- `npm test` reaches the new checks and they pass; the suite still has the existing local Windows Python alias failures in `aif-distillation` helper tests (`Python was not found`).

lee-to (approved):
Thanks for the follow-up. I re-checked the latest commit and the review gaps are addressed: the remaining user-facing templates are now explicitly structure-only and rendered in ui_language, and the /aif-rules config matrix/test contract now matches the limited writer behavior. CI is green. Approved.

PR #125, opened 2026-06-01
feat(dispatch): add cross-agent dispatch orchestration and support for chatgpt/openrouter

symonovserhii:
Description                                                                                                                                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                                                                                                                                 
  This PR implements the full Cross-Agent Dispatch (orchestration) spec and adds native support for ChatGPT CLI and OpenRouter CLI as both built-in agents and dispatch targets.
  
  ## Key Changes
  
  ### 1. Cross-Agent Dispatch Orchestration
  
  • Added command template resolution, command building (argv tokenizer/injection-safe), manual instruction rendering, and subprocess spawning in  src/core/dispatch.ts .
  • Registered  dispatchAgents  overrides and phase-specific configs in  src/core/config.ts .
  • Added the  ai-factory dispatch <phase>  CLI command in  src/cli/commands/dispatch.ts .
  • Integrated phase-dispatch boundary hooks into  skills/aif-loop/SKILL.md  (Step 4.0).
  • Created a comprehensive Bash integration test suite  scripts/test-dispatch.sh  covering fallback defaults, missing binaries, argv prompt-quoting, and error states.
  
  ### 2. Native Support for ChatGPT & OpenRouter
  
  • Registered  chatgpt  (ChatGPT CLI, config  .chatgpt/ ) and  openrouter  (OpenRouter CLI, config  .openrouter/ ) in  src/core/agents.ts  so users can initialize them via  ai-factory init .
  • Configured built-in dispatch runner specs in  src/core/dispatch-registry.ts :
      •  chatgpt-cli :  chatgpt --model {model} -p {prompt}  (defaults to  gpt-4o )
      •  openrouter-cli :  openrouter --model {model} -p {prompt}  (defaults to  anthropic/claude-3.5-sonnet )
  • Updated  cleanup-blocked-skill.py  whitelist and validation tests for  .chatgpt/skills  and  .openrouter/skills .
  • Added documentation across  getting-started.md ,  config-reference.md , and  README.md .
  
  ## Verification Results
  
  • Verified typescript compilation:  npm run build  compiles with 0 errors.
  • Verified unused exports/locals:  npm run lint  passes with no issues.
  • Verified test suites:
      •  npm run test:dispatch  (12/12 passed)
      •  npm run test:cleanup-helper  (30/30 passed)
      •  npm run test:extensions  (passed)
      •  npm test  (127/127 passed cleanly under redirected stdin)


ichinya:
## Code review

PR #125 adds cross-agent dispatch orchestration for `ai-factory dispatch <phase>`, new built-in dispatch runners for Claude/Gemini/Antigravity/ChatGPT/OpenRouter, config fields for `dispatch` / `phases` / `dispatchAgents`, `aif-loop` dispatch guidance, and a dedicated `test:dispatch` CI step. The core direction is reasonable, and the auto path avoids shell execution, but I would not merge/deploy this yet. I still found one runtime blocker in the manual dispatch path, and the branch needs a rebase against current `2.x` before CI can be trusted.

Current PR head SHA was not available from the accessible PR metadata; I reviewed the current `refs/pull/125/head` files.

### What looks good

* The auto execution path uses `execFileSync(file, args)` rather than a shell, which is the right baseline for avoiding command injection in spawned dispatch agents.
* The new resolver has safe no-op behavior for projects without dispatch config: no config / no phase entry exits 0 and lets the loop continue in-process.
* Dispatch test coverage is not only added as `scripts/test-dispatch.sh`; the PR also wires it into `package.json` as `npm run test:dispatch` and into the CI workflow as a `Dispatch tests` step.
* The tests cover important baseline paths: no config, missing phase entry, unknown phase, unknown agent, auto missing binary, `--print`, prompt argv preservation in auto mode, and the new `implement` / `review` phases.

### Must fix

1. [P1] Manual dispatch prints a non-copy-safe command for prompts with spaces or shell metacharacters — `src/core/dispatch.ts`

   `formatManualInstructions()` builds the displayed command with:

   ```ts
   const shown = [b.file, ...b.args].join(' ');
   ```

   This loses argv boundaries. The default prompts generated by `dispatchCommand()` already contain spaces, for example `Run the plan phase of the aif-loop.`. In manual mode, the printed command becomes a shell-looking command where the prompt is split into many shell words. A user copying the manual instruction will not run the same argv that auto mode would run. If the prompt contains quotes, `$()`, backticks, semicolons, newlines, or `&`, the manual copy/paste surface can also become misleading or unsafe even though the auto path is shell-safe.

   Broken scenario:

   ```bash
   ai-factory dispatch plan --prompt 'do the plan; echo leaked' --print
   ```

   The auto path would pass that as one argv item after `-p`; the manual instructions render it as raw shell text.

   Expected fix:

   * Render manual commands with explicit shell escaping for every argv element, or stop presenting them as copy/paste shell commands and instead print a command/args block that preserves argv boundaries.
   * Add a regression test where `--prompt` contains spaces, quotes, `$()`, semicolon, and newline-like content, and assert that the manual output is either safely shell-quoted or explicitly non-executable structured argv.
   * Keep the existing auto-mode argv assertion; add the equivalent manual-mode assertion so the two modes do not drift.

1. [P1] Do not merge until this branch is rebased onto current `2.x` and CI is rerun

   The repository default branch is `2.x`, and the current base `package.json` reports version `2.14.0`, while the PR head `package.json` reports `2.13.2`. This PR also touches `package.json` and `.github/workflows/ci.yml`, so stale-base merge risk is not theoretical: the package metadata and workflow/test gates need to be resolved against current base before review/CI status is meaningful.

   Acceptable options:

   * rebase onto current `2.x`, keep the current package version state, and rerun CI;
   * if the version rollback is intentional, document that explicitly in the PR body and update the lockfile/version policy consistently;
   * do not merge based on stale green checks from before the rebase.

### Should fix

1. [P2] Invalid dispatch config is silently normalized into a no-op or manual fallback

   `loadConfig()` drops unknown phase keys, drops invalid phase entries, and normalizes any non-`auto` `dispatch` value to `manual`. Safe fallback is good, but silent fallback is poor diagnostics for an opt-in orchestration feature. A typo like `"dispatch": "atuo"` or `"phases": { "plna": ... }` will look configured to the user but will not dispatch as intended.

   Expected fix or follow-up:

   * warn on unknown `dispatch` values and unknown phase keys;
   * keep safe fallback to `manual` / no-op;
   * add tests for invalid mode, typo phase key, missing `agent`, and empty model.

2. [P2] Extension runtime resolution is not hydrated in `dispatchCommand()`

   `resolveAgentRunSpec()` can resolve an installed runtime through `findAgentConfig(agentId)`, but `dispatchCommand()` only calls `loadConfig(process.cwd())`; it does not hydrate the project extension runtime registry first. In a fresh CLI process, an extension-provided runtime referenced from `phases.<phase>.agent` can be reported as unknown unless the user also defines it manually under `dispatchAgents`.

   Expected fix or follow-up:

   * call `hydrateProjectAgentRegistry()` with installed extension names before resolving dispatch;
   * add a smoke test with an installed extension runtime whose `skillsCliAgent` maps to a dispatch runner, and verify `ai-factory dispatch <phase>` resolves it.

3. [P2] Public config/docs contract for dispatch should be documented outside `aif-loop`

   The PR introduces public `.ai-factory.json` fields: `dispatch`, `phases`, and `dispatchAgents`. The skill text explains how `/aif-loop` calls the CLI, but users still need a stable config reference and examples for manual vs auto mode, custom runner templates, model fallback, and safe prompt handling.

   Expected fix or follow-up:

   * add a docs/configuration section or reference snippet for the new fields;
   * document that command templates are argv templates, not shell scripts;
   * document built-in ids separately from install runtime ids: `claude-code`, `gemini-cli`, `antigravity`, `chatgpt-cli`, `openrouter-cli` vs `claude`, `gemini`, `chatgpt`, `openrouter`.

### Nits

None.

### Context gates

* Architecture: warn — core/CLI separation is reasonable, but dispatch resolution should hydrate extension runtime definitions before claiming runtime-aware resolution.
* Package CHECKLIST: n/a — I did not find an accessible repo/package checklist in the inspected paths.
* Docs / skill contract sync: warn — `aif-loop` is updated, but the new `.ai-factory.json` dispatch contract needs a user-facing config reference.
* API/web/client sync: n/a — no web/client/API surface observed.
* Security surfaces: error — auto mode avoids shell execution, but manual mode renders unsafe/non-copy-safe shell text.
* Feature flag / risk gating: pass — behavior is opt-in through `.ai-factory.json`; auto spawning only happens with `dispatch: "auto"`.
* Env / Turbo / process plumbing: n/a — no env/Turbo changes observed.
* Managed assets / ownership: pass — new ChatGPT/OpenRouter runtime registration appears to use the existing install/managed-skill path.
* Migration / backward compatibility: warn — old configs no-op, but invalid new dispatch configs silently degrade.
* Dependency / lockfile / production install: warn — no new runtime dependency is added, but package metadata is stale relative to current base.
* PR size / decomposition: warn — dispatch core, new agent registrations, skill contract, tests, and CI are all in one PR; still mostly one coherent feature.
* Stack / merge order / base freshness: error — rebase required before merge because current base and PR head package/workflow state differ.
* CI / audit: warn — the PR head workflow includes `Dispatch tests`, but actual PR check status was not visible from this review context.

### Verification

* Checked: PR title/body search result, repository default branch, PR-head files under `refs/pull/125/head`, current base files under `2.x`, dispatch core, dispatch registry, CLI command registration, config normalization, agent registry changes, `aif-loop` Step 4.0, package scripts, dispatch test script, and CI workflow.
* Ran locally: not run; I reviewed source through GitHub file refs only.
* CI: workflow definition includes build, lint, existing smoke tests, and the new dispatch test step; current run/check status was not visible.
* Not verified: exact head SHA, author, labels, review comments, linked issues, mergeability, required-check conclusions, and exact behind/ahead count.

Review rubric applied from the uploaded reviewer instructions. 

### Verdict

Verdict: REQUEST_CHANGES until manual dispatch rendering is made argv-safe/copy-safe and the PR is rebased onto current `2.x` with CI rerun.

ichinya:
## Verdict: REQUEST_CHANGES

PR as a whole is moving in the right direction: a separate `ai-factory dispatch <phase>` has been added, the config has been expanded for `dispatch` /`phases` / `dispatchAgents`, auto-path launches the external CLI via `execFileSync(file, args)`without shell, and CI gets a separate `test:dispatch` step. It's a good base. But I would not fix the current version: there is one merge-gate P0 and several P1, which can break existing projects or make dispatch unsafe/unpredictable. The PR is also very wide: 36 files, +2088/-61, and along with dispatch, it changes update/config ownership, scanner, docs, and skill behavior.

## What looks good

Auto-dispatch does not collect the shell string, but calls `execFileSync(file, args, {stdio: 'inherit'})`, which is correct for basic protection against shell injection.

A new `test:dispatch` has been added to the `package.json`, and CI runs build, lint, basic smoke tests, cleanup tests, and dispatch tests.

`loadConfig()`it really starts saving new top-level dispatch fields during normalization. `ai-factory.json`, which was an important part of the backward-compatible round-trip.

---

# P0 — Blocker

## [P0] PR is not mergeable now

GitHub metadata shows `state: open`, `merged: false`, `mergeable: false'. Before review/merge, this should be considered a hard gate: the branch needs to be rebased/conflicts resolved and CI driven again.

**Why it matters:** even if the code looks correct in diff, the current PR cannot be safely merged into `2.x`, and the results of the old local checks from the PR body are not a sufficient merge signal.

**Expected fix:** rebase/merge current `2.x`, resolve conflicts, rerun full CI.

---

# P1 — Must fix before merge

## [P1] `ai-factory init` will lose the dispatch configuration when starting again

'AiFactoryConfig` now officially contains `dispatch`, `phases`, `dispatchAgents'.  But `initCommand()`, when saving the configuration, creates a new object only from `version`, `agents`, `extensions` and does not transfer dispatch fields from `existingConfig'.

**Broken scenario:**

1. The user has set up the `.ai-factory'.json`:

```json
{
  "dispatch": "manual",
  "phases": {
    "implement": { "agent": "antigravity" }
  },
  "dispatchAgents": {
    "custom": { "command": "..." }
  }
}
```

2. Then runs `ai-factory init` again, for example to add `chatgpt' or `openrouter'.
3. Dispatch-the settings disappear from the `.ai-factory'.json`.

**Why it matters:** this is a silent data loss for the new public config surface. The user does not expect that repeated init will delete the orchestra settings.

**Expected fix:**

* With `saveConfig()` in `initCommand()`, save existing `dispatch`, `phases', `dispatchAgents'.
* It is better to build the new config as a controlled merge on top of the `existingConfig` rather than as a completely new object.
* Add regression test: existing `.ai-factory.json` with dispatch fields → rerun `init` → fields preserved.

---

## [P1] Manual dispatch prints an unsafe/copy-unsafe shell-like command

`buildAgentCommand()` first splits the template by whitespace, then substitutes `{prompt}` as one argv element for auto-mode. This is normal for `execFileSync'. But `formatManualInstructions()` then does `[b.file, ...b.args].join(' ')`, losing argv boundaries.

**Broken scenario:**

```bash
ai-factory dispatch plan --prompt 'do the plan; echo leaked' --print
```

Auto-mode would pass prompt as one argv item after the `-p`. Manual-mode prints a raw shell-looking string with spaces, `;`, `$()`, backticks, quotes, or newline will be interpreted by the user/shell differently than auto-mode.

**Why it matters:** PR declares injection-safe command building, but manual path becomes unsafe copy/paste surface. This is especially bad because `manual` is the default mode.

**Expected fix:**

* Or shell-escape each argv element when rendering manual command.
* Or do not show the command as a shell command, but print a structured argv block, for example:

```text
Command: claude
Args:
  --model
  sonnet
  -p
  Run the plan phase of the aif-loop.
```

* Add tests for prompts with spaces, quotes, semicolon, `$()`, backticks, and newline.

---

## [P1] `update` prompt counts all missing base skills as “new skills” and selects them all by default

`promptForNewSkills()` builds candidates like any `availableSkill' that at least one agent does not have. These are not “new skills from the new version of the package”, but in general all the base skills that the user once intentionally did not install.  In this case, checkbox choices are created with `checked: true'.  Then `updateSkills()' actually installs all the selected `newSkillsToInstall'.

**Broken scenario:**

The user made a minimal project setup with `--skills aif` or a small set of skills. At the next interactive ai-factory update, he will be shown “New base skills available”, although these are just all the skills that he did not choose before. If he presses Enter on the default checkbox, the project will receive the entire missing base-skill set.

**Why it matters:** routine update may unexpectedly expand the skill surface of the project. This is a violation of user intent and a potentially very noisy change in managed skill directories.

**Expected fix:**

* Do not count “new” all missing skills. We need a way to distinguish “appeared in the package after the last update” from “the user intentionally did not install”.
* Before such a state appears, it is better to make choices `checked: false`, change the wording to “Uninstalled base skills available”, and require explicit opt-in.
* Add a non-interactive flag, such as `--add-new-skills` / `--no-add-new-skills` if this is to be supported in CI.
* Add a regression test for a project with intentionally small skill set: `update` should not turn it into `all` by default.

---

## [P1] “Transparent no-op” dispatch is not transparent: no-config/no-phase are written in stderr

`dispatchCommand()` for `no-config` and `no-phase-entry' writes messages via `console.error`, although exit code remains 0. At the same time, `aif-loop` must now call `ai-factory dispatch <phase>` before each phase, and the absence of a block on stdout means “run in-process exactly as today".

**Why it matters:** for all projects without dispatch-config, the new loop will execute the dispatch command at each phase boundary. Even if stdout is empty, stderr can get into the Bash/tool output and look like an error.: “No cross-agent dispatch configured”. This is not “loop unchanged".

**Expected fix:**

* For `no-config` and `no-phase-entry', by default, do not print anything at all.
* If you need diagnostics, add `--verbose` or `--explain'.
* Tests should check stdout and stderr separately, not the merged output.

---

## [P1] Dispatch resolver is not hydrate's extension runtime registry

`dispatchCommand()` only does `loadConfig(process.cwd())` and immediately calls `resolvePhaseDispatch()'.  But extension runtimes get into the registry only through `hydrateProjectAgentRegistry()`, which downloads the installed extensions and registers them with `manifest.agents'.  `resolveAgentRunSpec()` explicitly relies on `findAgentConfig(AgentID)` for alias/runtime resolution.

**Broken scenario:**

The extension adds a runtime `my-agent` with a `skillsCliAgent' that maps to the builtin dispatch runner. In `.ai-factory.json`:

```json
{
  "extensions": [{ "name": "my-extension", "...": "..." }],
  "dispatch": "manual",
  "phases": {
    "plan": { "agent": "my-agent" }
  }
}
```

In a separate CLI process, the `ai-factory dispatch plan` may return an unknown-agent because the extension registry is empty.

**Expected fix:**

* In `dispatchCommand()` after `loadConfig()`, call `hydrateProjectAgentRegistry(ProjectDir, { extensionNames: config.extensions?.map(e => e.name) ?? [] })`.
* After hydrate, reuse the registry-aware resolution.
* Add smoke test with extension-provided runtime.

---

# P2 — Should fix / follow-up before release if possible

## [P2] Auto-dispatch may freeze forever, despite contract “Dispatch must NEVER block the loop”

'runAuto()` synchronously calls `execFileSync()` without timeout/abort strategy.  Skill contract says that non-zero dispatch must fallback, and “Dispatch must NEVER block the loop.”  But if the external CLI freezes, asks for auth, opens interactive prompt, or just takes a long time to complete, the loop won't reach fallback.

**Expected fix:** add configurable timeout, timeout error path, and test fake CLI that `sleeps' longer than the limit. Minimally, it is necessary to explicitly document what `auto` can block, but it is better to bring the code to the contract.

---

## [P2] Invalid dispatch config silently normalizes/drops user intent

`normalizeDispatchMode()` turns any non-`auto` mode into `manual`; `normalizePhases()`silently ignores unknown phase keys, missing `agent`, empty `agent`, invalid `model`; `normalizeDispatchAgents()` silently throws invalid entries.

**Why it matters:** typo like `"dispatch": "atuo"` or `"phases": { "plna": ... }` will look like a configured dispatch, but it won't actually work.

**Expected fix:** leave the safe fallback, but add warnings to the `ai-factory dispatch` / `ai-factory update' or a separate config validation path. Add tests for invalid mode, typo phase, missing agent, whitespace command/model.

---

## [P2] `dispatchAgents.command` looks like a shell template, but is implemented as a whitespace-only argv template

Docs describe the `command` as a template string with `{model}`, `{phase}`, `{prompt}`.  The implementation does `spec.command.trim().split(/\s+/)`, meaning quoted args, escaped spaces, paths with spaces, and shell-like quoting are not supported.

**Expected fix options:**

* Better: replace config shape with `{ "command": "claude", "args": ["--model", "{model}", "-p", "{prompt}"] }`.
* Or support the documented quoting parser.
* Or explicitly write in docs: this is not a shell, quotes have no special meaning, template is divided by whitespace.

---

## [P2] Dispatch tests do not check stdout/stderr contract

'run_cli()` combines stdout and stderr via `2>&1'.  But skill contract distinguishes between stdout manual block and no block on stdout.  Therefore, current tests may skip regressions where no-op writes stderr, manual block goes wrong, or auto-mode makes noise in stdout.

**Expected fix:** capture stdout/stderr separately. For no-config/no-phase assert: stdout empty, stderr empty. For unknown phase/agent: stderr has diagnostic, stdout is empty. For manual: stdout has block, stderr is empty.

---

## [P2] PR is too wide for a single merge unit

In addition to dispatch, PR is changing config ownership behavior, `update` prompting, security scanner, `aif-improve` language directives, docs extensions catalog, and symlink tests. Given the size of 36 changed files and 2088 additions, this increases the review risk.

**Expected fix:** either split into several PRS, or explicitly highlight in the PR body which changes are prerequisites for dispatch and which are opportunistic cleanup. Minimally, it is necessary to do without focused tests for each individual behavior change.

---

# P3 — Nits / documentation cleanup

## [P3] `AGENTS.md `still says “16 agents”

`chatgpt` and `openrouter` have been added to the registry.  But `AGENTS.md ` continues to write “16 agents” and does not include ChatGPT/OpenRouter in the list.

**Fix:** Update count and list.

---

## [P3] Docs examples use the old version `2.13.2`, although the package is now `2.15.0`

`package.json` in PR sets `"version": "2.15.0"`.  But the dispatch config example in docs uses `"version": "2.13.2"`.

**Fix:** either remove the `version' from the example, or replace it with the current/placeholder value.

---

## [P3] `package.json` keywords not updated for ChatGPT/OpenRouter

README/docs advertise new runtime targets, and the keywords list ends with existing agent keywords and does not include `chatgpt`, `openrouter', `chatgpt-cli`, `openrouter-cli'.

**Fix:** Add keywords or deliberately leave unchanged if keywords are not supported as a discovery surface.

---

## [P3] `scripts/test-dispatch.sh `saves the `ORIGINAL_PATH`, but does not use it

In the test script, `ORIGINAL_PATH="$PATH"` is set, then the PATH is changed, but the variable is not used.

**Fix:** delete the variable or restore the PATH in cleanup.

---

# Suggested additional tests

1. `init` preserves `dispatch`, `phases`, `dispatchAgents` on rerun.
2. Manual dispatch output is shell-safe or structured argv-safe for prompt containing spaces, quotes, semicolon, `$()`, backticks, newline.
3. No dispatch config produces exit 0 with empty stdout and empty stderr.
4. Phase configured through extension-provided runtime resolves after registry hydration.
5. Interactive update on intentionally minimal skill set does not default-install all missing base skills.
6. Auto dispatch timeout/fallback with fake CLI that hangs.

---

## Final recommendation

**Request changes.** After the rebase/mergeability fix, I would block merge for at least P1 items: preserving dispatch on `init`, safe manual rendering, corrected update prompt semantics, truly silent no-op dispatch, and extension registry hydration. P2 can be partially removed by follow-ups, but it is better to close timeout/stdout-stderr tests before release, because they directly relate to the public dispatch contract.

ichinya (changes requested):
After fixing the rebase/merge option, I blocked the merging of several elements in P1: saving sending at startup, secure manual rendering, fixed update request semantics, really silent sending without operations, and updating the extension registry. P2 may need a follow-up check, without timeout/stdout-stderr tests, which may be closed before release, after which it will be sent to send the contract.

PR #132, opened 2026-06-21
refine architecture structure

DAGpro:


lee-to:
Review findings:

1. `skills/aif-architecture/references/architecture.md:23` still says Explicit Architecture has only two organization variants, but this PR adds and exposes the third `explicit-flat` variant. That makes the architecture reference internally inconsistent and may lead `/aif-architecture` to omit or mis-rank the flat variant during recommendation. Please update the decision/reference guidance to include the flat variant.

2. `skills/aif-architecture/references/architecture.md:73-74` duplicates `**Migration path to Explicit Architecture:**`. Looks accidental and should be removed.

3. `skills/aif-architecture/references/architecture.md` has trailing whitespace in the new tree examples. `git diff --check origin/2.x...pr-132-review` fails on these lines, so please strip trailing whitespace before merge.

Verification:
- `npm test` passes: 131 passed, 0 failed, 11 warnings.
- `git diff --check origin/2.x...pr-132-review` fails due to trailing whitespace.

lee-to:
Thanks, the previous review items look addressed now:

- `explicit-flat` is included in the subvariants note.
- The duplicated `Migration path to Explicit Architecture` heading is gone.
- `git diff --check origin/2.x...origin/pr/132` is clean.

I found two remaining issues:

1. `skills/aif-architecture/SKILL.md:79` says that when an architecture is provided via arguments, the skill should “skip to Step 2”. That bypasses the newly added Step 1.5 codebase alignment check. For explicit choices like `/aif-architecture explicit-layers` or `/aif-architecture explicit-flat`, we still need to compare the selected ideal structure with the existing codebase. Please change this to skip only the recommendation step and then continue to Step 1.5.

2. `skills/aif-architecture/references/architecture.md` now shows Structured Modules using `Infrastructure/`, but the dependency rules/key principles still reference `shared/` (`Modules depend on shared/`, `Shared is Minimal`). This makes the reference internally inconsistent and can lead generated architecture docs to mix both terms. Please align those rules with the new `Infrastructure/` terminology or restore `shared/` consistently in the example structure.

lee-to:
Thanks, the two previous items are addressed now:

- Architecture arguments now proceed to Step 1.5 instead of jumping straight to Step 2.
- Structured Modules terminology is aligned with `Infrastructure/` in the dependency rules and principles.
- `git diff --check origin/2.x...origin/pr/132` is clean, and the GitHub `test` check is green.

I only see two minor consistency follow-ups:

1. `skills/aif-architecture/references/architecture.md:23` says that “the matrix row shows the range across variants”, but the table does not show ranges; it uses a separate `Explicit Architecture (Vertical Slice)` column. Please either adjust the wording to say that variants with different scores may get a separate matrix column, or change the matrix to actually show ranges.

2. `docs/skills.md:283` notes that `clean`, `ddd`, and `monolith` are legacy aliases, but `skills/aif-architecture/SKILL.md` also documents `vertical` as a legacy alias. Please include `vertical` in the docs note so the user-facing docs match the skill instructions.

PR #134, opened 2026-06-23
Add QA check execution skill

lee-to:
## Why

AI Factory can already generate QA context with /aif-qa: a change summary, a test plan, and concrete test cases. What was missing was an execution step that turns those test cases into an auditable QA result. The new /aif-qa-check skill closes that gap by recording which scenarios actually passed, which failed, and why.

This is useful for two workflows:

- Human QA: guide a tester through one case at a time and persist the result after every answer.
- Agent QA: let an agent execute the same cases through live browser automation when the runtime provides a Browser integration or Playwright MCP.

## How it works

/aif-qa-check reads the existing /aif-qa artifacts from paths.qa/<branch-slug>/, especially test-cases.md, and writes a separate qa-check.md execution artifact. It intentionally keeps change-summary.md, test-plan.md, and test-cases.md read-only so generated QA design and QA execution results do not overwrite each other.

The skill supports two modes:

- human: shows exactly one TC-NNN case at a time, asks the user whether it works, checks passed cases, and records failed comments verbatim from the user. For Russian UI language, the required summary question is: "Протестируйте и ответьте работает или нет".
- agent: requires live browser testing. It prefers the in-app Browser capability, falls back to Playwright MCP, and stops before writing results if neither is available. Static inspection or shell-only checks are not enough to mark a case passed.

The result is saved incrementally to qa-check.md after every case, with checkboxes, status, comments, and browser evidence where available.

## Included changes

- Add skills/aif-qa-check with a config-aware, stack-free workflow.
- Add templates/QA-CHECK.md for execution result artifacts.
- Add aif-qa-check to CLI skill hints, AGENTS.md, workflow docs, skill docs, and config reference tables.
- Add scripts/test-aif-qa-check.sh and include it in npm test.

## Verification

- npm run build
- bash skills/aif-skill-generator/scripts/validate.sh skills/aif-qa-check
- python3 skills/aif-skill-generator/scripts/security-scan.py skills/aif-qa-check
- bash scripts/test-aif-qa-check.sh
- npm test

PR #135, opened 2026-06-23
Require research context for research-backed plans

lee-to:
## Why

Research-backed plans currently preserve the Active Summary in the plan, but downstream skills did not have a strict contract to re-read `RESEARCH.md` when executing, refining, or verifying those plans. That could cause agents to rely only on the copied summary and miss constraints, decisions, or rationale that remain in the research artifact.

This change makes the research handoff explicit and durable: when `/aif-plan` uses research, it must record `Research Context` with a `Source:` reference, and downstream plan consumers must re-read the resolved `paths.research` artifact before acting on the plan.

## What changed

- Require `/aif-plan` to include `Research Context` and a source reference when research influenced the plan.
- Require `/aif-implement`, `/aif-verify`, and `/aif-improve` to read `paths.research` when a plan references `RESEARCH.md`.
- Extend `/aif-fix` to resolve `paths.research` and re-read it when an existing fix plan references research.
- Update the canonical plan template, docs, config reference, and AGENTS guide to match the new contract.

## Verification

- Ran `git diff --check`.


ichinya:
Verdict: Request changes

Two contract blockers make the merge unsafe.

[P1] The plan can get requirements from someone else's newer research

/aif-explore replaces the contents of the single Active Summary, keeping the old information only in Sessions. The new /aif-implement and /aif-verify instructions then read the current Active Summary and consider it to be the plan's requirements.

Reproducible scenario:

RESEARCH.md contains topic A.
/aif-plan creates plan A and copies the current summary there.
Later, /aif-explore replaces the Active Summary with topic B.
Running plan A through /aif-implement mixes in requirements B.
/aif-verify can verify implementation A against requirements B.

 This can lead to implementation of an incorrect scope or a false verification failure.

 Expected fix: consider Research Context copied into the plan as a committed set of requirements. Source needs a stable revision - timestamp, hash or session-specific reference. If the current revision does not match, a drift warning should be issued instead of silently replacing the plan requirements.

Regression test: create a plan from topic A, replace the Active Summary with topic B, then execute implement/verify for plan A. The requirements of B should not be applied; an explicit drift warning should appear.

[P1] /aif-improve does not save or create a research linkage

Skill now reads research for an already marked plan and can read it for a regular plan when it considers it relevant.

But when applying refinements, full regeneration is allowed via Write; only the completed checkbox and filename prefix are explicitly saved. Saving or adding Research Context and Source is not required.

Because of this:

a research-backed plan may lose its reference after extensive refinement;
a regular plan may receive requirements from research but not be marked as research-backed;
a subsequent /aif-implement or /aif-verify will skip the mandatory research reading.

Expected fix: add an invariant to Step 6.5:

preserve existing Research Context and Source on any rewrite;
if research has influenced refinement, add or update these fields even for a previously untagged plan;
preserve committed revision of source.

Regression tests: linked plan → extensive rewrite → linkage preserved; unlinked plan → research-informed refinement → linkage added.

[P2] New FIX_PLAN cannot actually become research-backed by the regular way

/aif-fix resolves paths.research and adds an optional section to the template, but only reads DESCRIPTION, ARCHITECTURE, rules, and skill-context when creating a new plan. There is no reading of research before creating a FIX_PLAN.

 You should either read resolved research when creating a plan-first fix and transfer the snapshot/source, or remove the promise from the template that the regular workflow does not implement.

Merge gates
Head 8827a2b is 12 commits behind the current 2.x; the intermediate changes include docs/skills.md, which was also modified by this PR. An update/rebase and a re-CI are required.
The current CI completed successfully, but it does not verify the described lifecycle scenarios.
The PR is open, not draft, and consists of one commit and 10 files; the title/body match the diff. There are no previous comments or review findings.

lee-to:
Addressed the review findings:

- Research-backed plans now treat embedded `Research Context` as the committed requirements snapshot.
- `/aif-implement`, `/aif-verify`, `/aif-improve`, and `/aif-fix` now check live research only for revision drift and emit `WARN [research-drift]` instead of applying newer Active Summary content silently.
- `/aif-improve` preserves/adds research linkage on rewrites, including revision metadata.
- `/aif-fix` plan-first mode now reads relevant research before creating `FIX_PLAN.md` and records the committed source revision.
- Added a regression check in `scripts/test-skills.sh`.

Also rebased the branch on the current `origin/2.x` and force-pushed with lease.

Validation:
- `git diff --check`
- `npm test` -> 132 passed, 0 failed

ichinya:
## Verdict: **Request changes**

 I rechecked the current PR. The old uploaded diff/md files are no longer available in the session, so I checked the current GitHub head directly. If you need to compare with locally saved files, you will need to upload them again.

### What has changed since the last review

PR updated: now head 83eb2b78..., 2 commits, 12 files, 63 additions / 14 deletions; PR is open, not draft, mergeable.

Branch is now rebased: 2.x..83eb2b7 shows `ahead_by: 2`, `behind_by: 0`.

CI on the new head was successful.

Previous comments are mostly covered by the contract text:

* `/aif-implement` now explicitly uses embedded `Research Context` as a committed scope and live `RESEARCH.md` only for drift-check with `WARN [research-drift]`.
* `/aif-improve` now must save `Research Context`, add it to the unlinked plan for research-informed refinement, and warn about drift.
* `/aif-fix` plan-first now reads `paths.research` before creating a fix plan and adds the source revision if the research is relevant.

### [P1] The new contract requires `SHA256`, but `/aif-plan` and `/aif-improve` cannot calculate it

The new text requires `/aif-plan` to record a stable revision marker: `Updated:` timestamp plus `SHA256:` from the copied Active Summary. In Step 5, this is already formalized as a mandatory source line with `SHA256: <sha256 of copied Active Summary>`.

But `/aif-plan`'s `allowed-tools` only allow `Bash(git *)`, `Bash(cd *)`, `Bash(cp *)`, `Bash(mkdir *)`, `Bash(basename *)`; there is no `sha256sum`, `shasum`, `python`, `node`, `openssl` or other way to calculate SHA-256.

The same problem with `/aif-improve`: during research-informed refinement, it should add `Source: ... SHA256: <sha256 of copied Active Summary>`, but its `allowed-tools` are limited to `Bash(git *)` and do not allow the calculation of SHA-256.

 This makes the new contract impossible to execute in a regular run. The agent will either leave the placeholder, create a hash, or violate the tool allowlist. Drift-check will then become unreliable.

 **Fix:** choose one of the following options:

1. expand `allowed-tools` for `/aif-plan` and `/aif-improve` to an explicit, portable hash path, such as `Bash(sha256sum *)` plus `Bash(shasum -a 256 *)`, and add instructions on how to normalize the Active Summary before hashing;
2. or remove the mandatory `SHA256` from the producer/consumer contract and use only the `Read` `Updated:` marker;
3. or make the hash optional: `Updated:` is required, `SHA256:` is required only if the hash tool is allowed.

 It is also worth updating `scripts/test-skills.sh`: the current regression check checks for the presence of lines about `SHA256`, but does not check that the skills that are required to write the hash have a valid tool for calculating it.

### Merge gates

Do not merge until P1 is fixed. After the change, you need to repeat the CI. The rest of the previous blockers looks closed at the current diff level.

lee-to:
Fixed the P1 hash contract issue in 71132e6.

What changed:
- Added explicit SHA-256 calculation tools to `/aif-plan` and `/aif-improve`: `shasum -a 256` and `sha256sum`.
- Documented the normalization procedure for the copied Active Summary before hashing, so the recorded `SHA256:` is reproducible.
- Added a regression check in `scripts/test-skills.sh` to ensure the producer skills keep both the allowed tools and normalization instructions.

Validation:
- `npm test` passed: 133/133.

ichinya:
CI on the new head was successful.

 The previous P1 about the inability to calculate `SHA256` in `/aif-plan` and `/aif-improve` is mostly closed: both skills have received `shasum -a 256` / `sha256sum`, and the rules for normalizing the copied Active Summary before hashing have been added.

### [P1] `/aif-plan` still makes a research-backed plan based on the existence of `RESEARCH.md`, rather than the influence of research

In an early step, `/aif-plan` formulates a correct condition: if the research content affects the plan, then the plan should receive the `## Research Context` and source revision.

 However, Step 5 remains without a relevance gate: **if the resolved research path exists**, the skill must copy the `Active Summary`, write the `Source`, and then read the copied `Research Context` committed requirements snapshot.

 This is still dangerous:

1. The project still has the `.ai-factory/RESEARCH.md` from the old A theme.
2. The user runs `/aif-plan full <new unrelated task B>`.
3. Step 5 still copies the Active Summary of topic A.
4. The new plan B becomes research-backed and downstream will consider the old topic A to be part of the committed scope.

 This is not just noise in the plan: after the current PR, the embedded `Research Context` has intentionally become a source of requirements for implementation / verification. Therefore, stale or unrelated research may be included in the scope of the new plan.

 **Fix:** In Step 5, replace the condition `If the resolved research path exists:` with `If research content influenced this plan:`. If research was read but found to be irrelevant, `## Research Context` should not be added. It is advisable to add a regression check: an existing `RESEARCH.md` with topic A + an explicit `/aif-plan` for unrelated topic B should not create `## Research Context`.

### [P2] Hash path leaves the temporary file outside the ownership boundary

The new hash workflow says to write the normalized text to a temporary file under `.ai-factory/.tmp/`, then run `shasum`/`sha256sum`.

 But `/aif-plan` explicitly restricts ownership: the team only owns the plan files — the resolved fast plan path and files under `paths.plans`; other artifacts must be maintained by the owner commands.

The same problem exists in `/aif-improve`: hash instruction writes temp file under `.ai-factory/.tmp/`, and ownership allows to edit only active plan artifact; description, architecture, roadmap, rules, and research are read-only context.

 There is no instruction to delete this temp file now, and `/aif-plan` has no permitted `rm` in allowlist. Bottom line: a regular `/aif-plan` or `/aif-improve` can leave an extra untracked file in repository and break its own ownership boundary.

**Fix:** either calculate hash without file artifact, or write temp to explicitly allowed place and be sure to delete it before completion, adding safe allowlist for cleanup. Even better - fix it in test: producers that create temp hash input must have cleanup contract.

### Merge gate

Result: previous blocker about missing hash tool is closed, but PR still not worth merging because of unconditional research linkage in `/aif-plan` and temp-file side effect.

lee-to:
Addressed the review feedback in 1fd8139.

Changes:
- /aif-plan now tracks whether research actually influenced the plan and only writes `## Research Context` when it did. Existing stale or unrelated RESEARCH.md content is explicitly ignored for unrelated plan requests.
- /aif-plan and /aif-improve now compute the Active Summary SHA256 without writing temp files or repository artifacts.
- Updated the plan template/docs and added regression checks for the relevance gate and no-temp-artifact hash contract.

Verification: `npm test` passes (133 passed, 0 failed; existing validator warnings only).

ichinya:
The contract is well designed, and this is a significant improvement over the previous state. But there's a real consistency gap in this, as well as an outdated rebase stage, so I can't quite agree (LGTM).

### Response pass

- Fixed: previous P1 "plan becomes research-backed only based on existence `RESEARCH.md `". Step 5 `/aif-plan' now restricts the `## Research Context` through "if the content of the research influenced this plan" with an explicit rule for unrelated topic B, and the `docs/skills.md' + template reflect this. The 'research_influenced_plan` gate is in place.
- Fixed: previous P2 "hashing path leaves temporary file out of ownership". '/aif-plan` and `/aif-improve` now calculate the digest via stdin/inline `shasum -a 256`/`sha256sum` without a writable artifact, and the regression test checks for the absence of `.ai-factory/.tmp/` and `<tmp-file>`.
- A new blocker discovered: `/aif-fix` has been added to the same SHA256 contract, but has not received normalization/hashing instructions and is not covered by the regression test for hashing tools. See below P1.

### What looks good

- Fixed handoff of research results is a valid model: embedded snapshot = fixed requirements, live `RESEARCH.md ` = drift check only, and `WARN [research-drift]` preserves the old viewport instead of silently expanding it. This directly eliminates the "implementation of theme B versus theme A" error class raised in the first review.
- The normalization steps in `/aif-plan` and `/aif-improve` are specific and reproducible (exclude `Source:` line and comments, keep the line order, LF, end with exactly one final line break, take the first field from `shasum`/`sha256sum`). The cross-platform double `shasum -a 256`/`sha256sum' is valid for macOS/Linux.
- The changes in `allowed-tools` are consistent for the two skills that need them (`aif-plan`, `aif-improve`), and `/aif-fix` uses unfiltered `Bash', so the tools are available there as well. No new contract is unenforceable due to the resolution of the instruments.
- Documentation synchronization is active: `docs/skills.md', `docs/workflow.md', `docs/plan-files.md`, `docs/config-reference.md` (the reading matrix for `/aif-verify` and `/aif-fix` now includes `paths.research`) and `AGENTS.md `coincide with skill changes. The `## Research Context` template in `TASK-FORMAT.md ` marked with a revision marker and the comment "snapshot of fixed requirements".
- The ownership contract is preserved: producers write only plan files (`/aif-improve` explicitly calls research/description/architecture/roadmap read-only), and consumers remain in read-only mode for `RESEARCH.md `.

### Must fix

1. [P1] `/aif-fix` follows the same SHA256 contract, but has no normalization/hashing instructions and is not validated by the hashing regression test — `skills/aif-fix/SKILL.md`

   The `/aif-fix' plan-first mode now instructs the agent to write `Source: <resolved research path> (Active Summary, Updated: <research Updated timestamp>, SHA256: <sha256 of copied Active Summary>)` (lines 193 and 250 in head PR). This is identical to the producer's commitment in `/aif-plan`/`/aif-improve'. But, unlike them, `/aif-fix` **does not contain a single word ** on how to normalize the copied Active Summary before hashing, how to transfer it to `shasum`/`sha256sum` or the rule "exactly one final line break". Confirmed: `grep` by `shasum|sha256sum|normalize|final newline|stdin|inline shell` in `skills/aif-fix/SKILL.md` on 1fd8139 returns NONE.

   Why this is important: without a normalization contract, the `SHA256` value recorded by the patch plan is not reproducible, so the drift-check from the consumer is not reliable — exactly the kind of failure that this PR exists to prevent, only for a separate code path. This cannot happen with `/aif-plan`/`/aif-improve` because they have instructions, so this is an internally agreed omission, not an impossible contract.

   What makes the situation worse: a new regression test in `scripts/test-skills.sh ` checks for normalization/hashing tools only in `AIF_PLAN_SKILL` and `AIF_IMPROVE_SKILL', not in `AIF_FIX_SKILL'. Therefore, the `npm test' (133/133) passes despite this gap — the test gives a false sense of security for exactly the surface that is not covered. This is especially important because `/aif-fix` has an unfiltered `Bash`, so the tools are *available*; only the instructions and the assertion in the test are missing.

   Expected fix:
   - Add the same normalization/hashing step to the `/aif-fix` plan-first mode as in `/aif-plan` Step 0 / `/aif-improve` 2.x (exclude `Source:` line and comments, line order, trim trailing spaces, LF, exactly one final line break, pass via stdin/inline to `shasum -a 256` or `sha256sum`, without a temporary file, take the first field).
   - Expand the hashing verification regression block in `scripts/test-skills.sh `to also assert the presence of normalization instructions and the absence of `.ai-factory/.tmp/` in `AIF_FIX_SKILL` so that the three producers adhere to the same verifiable standard.

   Expected regression/smoke test:
   - Basic verification path: create `RESEARCH.md ` for topic A, start the `/aif-fix` plan-first mode for the error related to A; received `FIX_PLAN.md ` must have a `## Research Context` with a `Source:` a string whose `SHA256:` corresponds to `shasum -a 256` of the normalized copied Active Summary.
   - Drift test: After creating a correction plan, replace the Active Summary in `RESEARCH.md `on topic B and run `/aif-fix` without arguments (execution); it should issue `WARN [research-drift]` and execute against embedded topic A, not topic B.

1. [P1] Do not merge with diverging databases; first perform a rebase on the current `origin/2.x` and restart CI

   `2.x` has moved forward: the comparison of `2.x...` 1fd8139 reports `ahead_by: 4, behind_by: 3, status: "diverged"`. The 3 commits that are not in this PR include entries in `AGENTS.md `, `docs/skills.md', `docs/workflow.md`, `docs/config-reference.md` and, most importantly, `scripts/test-skills.sh ` (qa-check merge #134 + related test scripts). This is a direct, not indirect overlap: PR rewrites the same regression domain `scripts/test-skills.sh ` and the same docs/AGENTS sections as the base branch. The captured baseRefOid is 19c7ff1, while the real `2.x` is 8479cf9, so the PR was created for an outdated database.

   Why this is important: The CI that passed (green) on 1fd8139 was started for a tree that is no longer something that would have been merged; a text merge with the modified `test-skills.sh ` may cause new regression blocks (added by this PR and #134) to both be present, but the order/logical logic was not tested together. The "clean" mergeStateStatus reflects the absence of a Git conflict, not that the combined test suite actually passes.

   Acceptable options:
   - perform a rebase on the current `origin/2.x' and resend, then run the `npm test` and CI again;
   - if there are text conflicts in `scripts/test-skills.sh `no, confirm anyway that the combined file contains both sets of regression checks and they do not overlap.;
   - a link to the evidence from the CI launch of the rebased head, not to this current green launch before merging.

### Should fix

1. [P2] Consumers accept `Source:`/`Reference:`/`any path or link` to `RESEARCH.md `as a signal, "research affects this plan," which is broader than the producer's gate

   '/aif-implementation`, `/aif-verify`, `/aif-improve` and `/aif-fix` are triggered on drift branches if the plan contains `Source:`/`Reference:`/`Reference:`/`any path or link` to `paths'.research`. This is good for the plans created by this PR. But this means that the existing plan, which has `Source: .ai-factory/RESEARCH.md (Active Summary)` without a revision marker (the old template format), now calls the branch "no revision marker → `WARN [research-drift]`" every time it is executed. This does not disrupt operations, but it is a false drift trigger for historically created plans, and it must be documented as an intentional one-time warning, or consumers must suppress the drift when `Source:` the line does not contain a revision marker at all. It's worth at least adding a note to the `docs/plan-files.md'.

2. [P2] The normalization step has been duplicated verbatim in `/aif-plan` and `/aif-improve` (and should now be in `/aif-fix`)

   The same normalization paragraph is copied between skills. As the set of producers grows, keeping three copies manually synchronized is exactly how regressions appear (as shown by the gap in `/aif-fix`). Consider extracting the normalization/hashing contract in `skills/aif-plan/references/RESEARCH-REVISION.md` (or the general `references/`) and pointing each producer to it, while preserving SKILL as a decision matrix. This is consistent with the own policy of the architecture of this repository — to keep SKILLS compact, and dense material in `references/'.

### Nits

- `docs/plan-files.md` example `Source:`still uses the hardcoded `2024-01-15 10:30`; acceptable as a literal in the example, but consider `YYYY-MM-DD HH:MM` to avoid the impression of a "properly formatted date" for readers who copy/paste.
- In `docs/skills.md` `/aif-plan`, the item is now readable "can read Active Summary as an optional context... Includes the Research Context only when the content of the study has been influenced," which is good; consider applying the same wording "only when the study has been influenced" to the description of the `/aif-fix` plan-first mode for consistency.

### Verification

- Tested: full differential on the head of 1fd8139 through `gr pr diff'; presence/absence of instructions on normalization and hashing in `AiF-plan`, `AiF-boost` and `AiF-fix on the head of PR through API; ads allowed-tools for all three producers; declarations of test variables (`AIF_FIX_SKILL', `AIF_VERIFY_SKILL') and the exact lines `grep command` of the new regression blocks in `scripts/test-skills.sh `; comparison of `2.x...1fd8139' (`ahead_by 4, behind_by 3, diverged`) and the list of files modified by `2.x` since the capture of the base (19c7ff1 → 8479cf9); ki status at the current level.
- Unexpectedly: the 'npm test' did not run on your system; updates 133/133 by the PR author, but nevertheless, the regression package does not open up the possibility for '/aif-fix'.
- Ci: "test" passed (1 m 44 s) on 1fd8139 — outdated/divergent head, so this green status is not proof of a unified state.
"At first: tactical overall change after the relocation to origin/2.x; it was not possible to determine whether there was a technical result in `scripts/test-skills.sh (these words refer to the conflict's activities, but the general order of action is not limited by space).

### Verdict

Verdict: REQUEST FOR CHANGES, the gap in the consistency of `/AiF-Fix` (P1 — normalization instructions + test extension) will not be eliminated, and the PR will not be rebased to the current origin/2.x` with a restart of Ci due to a divergent, overlapping base state. The previous blockers (relevance gateway, temporary file artifact) have been eliminated; as soon as these two items are completed, it will be approved with the specified clarifications P2 as subsequent tasks.

lee-to:
Thanks for the detailed review. Fixed and pushed in `ac697f9`.

What changed:
- Rebased the branch onto the current `origin/2.x` (`8479cf9`); the branch is now `0 behind / 5 ahead` against the base.
- Added the missing `/aif-fix` plan-first normalization and SHA256 instructions: exclude the `Source:` line and comments, preserve order, trim trailing spaces, use LF, end with exactly one final newline, feed stdin or inline shell input to `shasum -a 256` / `sha256sum`, take the first field, and do not write temporary artifacts.
- Expanded the regression block in `scripts/test-skills.sh` so the SHA256/temp-artifact guard covers `AIF_FIX_SKILL` in addition to `/aif-plan` and `/aif-improve`.
- Documented the legacy `Source:` / `Reference:` without revision metadata behavior as a compatibility `WARN [research-drift]` path in `docs/plan-files.md`.
- Updated the `/aif-fix` docs wording so `Research Context` is included only when research content influenced the fix plan, and changed the example timestamp placeholder to `YYYY-MM-DD HH:MM`.

Verification:
- `npm run build` passed locally.
- `npm test` passed locally: 135/135.
- GitHub CI `test` passed on the rebased head: https://github.com/lee-to/ai-factory/actions/runs/28298891047/job/83843603337

This should address both P1 blockers plus the documented P2/nit items from the review.

ichinya:
## PR #135 Review: Require research context for research-backed plans

**Overview**: PR establishes a strict contract for the transfer of research data between skills. It's a good idea to do a first-class citizen study in workflow.

### Positive moments

1. **Clear semantics** — `Research Context` as a committed requirements snapshot with `SHA256` for version control. This is correct: the plan should not "float" when the initial study is changed.

2. **Drift detection** — `WARN [research-drift]` instead of silent application. The agent will not pick up a new study without the user's knowledge.

3. **Hash Normalization** — detailed normalization instructions (LF, trim, exclude Source line). Important for reproducibility.

4. **Tests** — added extensive checks in `test-skills.sh `for the availability of a contract in all affected skills.

### Remarks

**1. Terminology of "committed requirements snapshot"**
The `aif-plan/SKILL.md:168` says "committed requirements snapshot", but it's not a git commit — it's an invariant inside the plan. It could have been clearer: "the plan's authoritative requirements copy" or "the plan-owned research copy". Minor.

**2. `allowed-tools` extension**
`aif-improve` got `Bash(mkdir *) Bash(shasum -a 256 *) Bash(sha256sum *)'. This is reasonable for hashing, but why `mkdir'? The diff does not show the use of `mkdir` in this skill.

**3. Backward compatibility**
The `docs/plan-files.md:174` describes the processing of legacy plans without revision markers:
> Plans created before revision markers were introduced may contain a Source: or Reference: line to RESEARCH.md without Updated: or SHA256: metadata.

This is good, but in SKILL.md However, there is no explicit indication of what exactly the fallback looks like. `aif-implementation/SKILL.md:473` says "emit WARN [research-drift]", but does not specify what to do if修订元元 is missing. It is worth adding: "treat as an unverified research link" (which is already available in plan-files.md , but not in the implementation).

**4. Consistency in `aif-fix`**
`aif-fix/SKILL.md:191-192`: hashing is described twice — in Step 2 and in the plan template. One description is enough, the second can be shortened to a cross-reference.

**5. No changes in the 'aif-loop`**
'/aif-loop` is mentioned in `config-reference.md `as a reader of `paths.research`, but there are no changes in `aif-loop/SKILL.md` in diff. If the loop creates plans via `/aif-plan`, it's OK. But if the loop reads the study directly, it's worth checking out.

### Bottom line

**Verdict: APPROVE with minor suggestions**

The contract is well thought out, the tests cover key scenarios. I recommend:
- Clarify the fallback for legacy plans in the 'aif-implementation`
- Remove unused `Bash(mkdir *)` from `aif-improve' if not planned
- Check the `aif-loop` for the need for similar changes

ichinya (approved):

PR #138, opened 2026-07-03
Preserve original aif-plan request in plan files

lee-to:
## Summary

- Add an `Original Request` contract to `aif-plan` so explicit user planning requests are stored verbatim in generated plan files.
- Omit `Original Request` only when a plan is created solely from `RESEARCH.md` without an explicit user request.
- Update the canonical plan template, workflow docs, skills reference, plan-files docs, and AGENTS.md to reflect the new plan-file contract.

## Motivation

During multiple `aif-improve` iterations, the starting task can drift or get lost. Preserving the original user request in the plan file gives downstream planning, improvement, implementation, and verification steps a stable reference point for the initial intent.

## Verification

- `git diff --check`


ichinya:
## PR Review

### P0 — Blocking

None.

### P1 — Should fix before merge

1. **Propagate the `Original Request` contract to `/aif-improve`.**

   The PR motivation is to prevent the initial task from drifting across multiple `/aif-improve` iterations, but the new contract is currently defined mostly in `/aif-plan` and documentation. `/aif-improve` still does not explicitly treat `## Original Request` as an immutable scope anchor, and its plan rewrite instructions do not require preserving that section verbatim.

   Please update `skills/aif-improve/SKILL.md` so that it:
   - reads `## Original Request` as the original user intent / scope anchor;
   - preserves `## Original Request` exactly on any plan edit or regeneration;
   - does not translate, summarize, normalize, or rewrite it even when `artifact_language` differs;
   - uses it when deciding whether a refinement is in scope or out of scope.

   Without this, the new section may be lost or rewritten in the exact workflow this PR is intended to protect.

2. **Add an explicit language-policy exception for `Original Request` in downstream plan rewrites.**

   `/aif-plan` correctly says that `Original Request` is raw user input and must not be translated. However, downstream skills such as `/aif-improve` generally require generated or updated plan artifact content to be written in `artifact_language`. That creates an ambiguity: a full plan rewrite could translate the original request despite the new contract.

   Please add an explicit exception wherever plan files may be rewritten: `## Original Request` is raw source input, not generated artifact prose, and must remain verbatim.

### P2 — Nice to have / clarity

1. **Move the `Original Request` docs out of the Roadmap Linkage section.**

   In `docs/plan-files.md`, the new `Original Request` paragraph appears under `## Roadmap Linkage (Optional)`. It is a separate plan-file contract, not a roadmap-specific concern. Consider giving it its own subsection, for example `## Original Request`, before `## Roadmap Linkage (Optional)`.

2. **Clarify what “after command mode/flags are removed” means.**

   The current parsing rule could be interpreted as removing any occurrence of `fast`, `full`, `--parallel`, etc. from the request text. It would be safer to state that only recognized command tokens are removed, not matching words inside the user’s actual request.

   It may also be worth specifying whether the preserved request is:
   - exact remaining text including internal whitespace;
   - trimmed only at the outer edges;
   - normalized in any other way.

3. **Optionally mention `Original Request` in `/aif-implement` and `/aif-verify`.**

   This is not strictly required for this PR, but it would make the contract more robust if implement and verify flows explicitly understood `## Original Request` as useful scope context, while still treating the task list and committed `Research Context` as the executable plan inputs.

lee-to:
Thanks for the review. I pushed a follow-up commit addressing the recommendations:

- Updated `/aif-improve` to treat `## Original Request` as the immutable original intent / scope anchor.
- Made plan rewrites preserve `## Original Request` exactly and exempt it from `language.artifacts` translation/normalization.
- Clarified `/aif-plan` parsing so only recognized command tokens in command positions are removed, with only outer whitespace trimmed.
- Moved the `Original Request` docs into their own section in `docs/plan-files.md`.
- Added downstream context notes for `/aif-implement` and `/aif-verify`.

Verification: `git diff --check` passes.

ichinya:
## PR Review

### Summary

Thanks for the follow-up. The updated PR addresses the previous review points:

- `/aif-improve` now treats `## Original Request` as the immutable original intent / scope anchor.
- Plan rewrites now preserve `## Original Request` verbatim and explicitly exempt it from `language.artifacts` translation/normalization.
- `/aif-plan` parsing now removes only recognized command tokens in command positions and preserves the remaining user request with only outer whitespace trimmed.
- `docs/plan-files.md` now has a dedicated `## Original Request` section instead of placing this contract under Roadmap Linkage.
- `/aif-implement` and `/aif-verify` now mention `Original Request` as useful original scope context while keeping the task list and committed `Research Context` as the executable inputs.

CI is green on the latest head.

### P0 — Blocking

None.

### P1 — Should fix before merge

None.

### P2 — Non-blocking nits

1. **Align the canonical task-format comment with the new parsing contract.**

   `skills/aif-plan/references/TASK-FORMAT.md` still says “Preserve the request verbatim after command mode/flags are removed”. The main `/aif-plan` instructions are now more precise: remove only recognized command tokens in command positions and trim only outer whitespace.

   This is not blocking because the authoritative `/aif-plan` flow is clear, but it would reduce future ambiguity if the template comment used the same wording.

2. **Consider making the `## Original Request` heading exception explicit in `/aif-plan` language policy.**

   `/aif-plan` now clearly says that the request body must not be translated, and the contract names the section as `## Original Request`. However, the general language policy still says generated plan artifact headings are translated when `artifact_language` is not `en`.

   This is likely resolved by the more specific later contract, but an explicit one-line exception near the language policy would make the intent harder to misread:
   `Exception: the section heading and body of ## Original Request are fixed raw-source structure and must not be translated.`

### Recommendation

Approve. The remaining items are minor consistency nits and do not block merge.

lee-to:
Thanks for the approval. I also pushed a small follow-up for the two non-blocking nits:

- Updated `skills/aif-plan/references/TASK-FORMAT.md` so the canonical template comment matches the precise parsing contract.
- Added an explicit `/aif-plan` language-policy exception that the `## Original Request` heading and body are fixed raw-source structure and must not be translated or rewritten.

Verification: `git diff --check` passes.

faugera:
OK

PR #149, opened 2026-08-16
feat(agents): add zcode agent support

povly:
Add ZCode as a built-in runtime with full skill and MCP support:

- Register zcode agent (configDir .zcode, skillsDir .zcode/skills)
- Write MCP servers to .zcode/config.json under mcp.servers using the strict stdio schema ({ type, command, args, env }) and remove them from the nested container on extension cleanup
- Add ZCodeTransformer: skills install unchanged into .zcode/skills/ and postInstall generates /aif-* slash-command stubs in .zcode/commands/ (skills: frontmatter auto-mounts the skill, $ARGUMENTS passes invocation arguments); managed stubs are marked and pruned when their skill disappears
- Document the ZCode runtime contract in the aif skill MCP matrix
- Update docs (getting-started, configuration), README, AGENTS.md (16 -> 17 agents), and package.json keywords

Verified: npm test 145/145; live ai-factory init --agents zcode run installs 29 skills, 29 command stubs, and a working .zcode/config.json; /aif appears in the ZCode / menu after session restart.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -9,8 +9,8 @@
 1. **CLI tool** (`ai-factory init/update/upgrade`) — installs skills and configures MCP
 2. **Built-in skills** (29 skills, all `aif-*` prefixed) — workflow commands for spec-driven development
 3. **Spec-driven workflow** — structured approach: plan → implement → commit
-4. **Multi-agent support** — 16 agents (Claude Code, Cursor, Windsurf, Roo Code, Kilo Code, Antigravity, OpenCode, Warp,
-   Zencoder, Codex CLI, Codex app, GitHub Copilot, Gemini CLI, Junie, Qwen Code, Universal)
+4. **Multi-agent support** — 17 agents (Claude Code, Cursor, Windsurf, Roo Code, Kilo Code, Antigravity, OpenCode, Warp,
+   Zencoder, ZCode, Codex CLI, Codex app, GitHub Copilot, Gemini CLI, Junie, Qwen Code, Universal)
 
 ## Project Structure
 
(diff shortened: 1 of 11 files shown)


ichinya:
## Review — changes requested

Reviewed at head 57578f1ba02c41e221bdaf71a80816f8495543b8.

### P1 — Existing ZCode workspace commands can be overwritten

`ZCodeTransformer.postInstall()` writes every generated stub unconditionally (`src/core/transformers/zcode.ts:52-55`). `.zcode/commands/*.md` is also ZCode's native workspace location for commands created or imported by the user.

Therefore an existing command such as `.zcode/commands/aif.md` is silently replaced during init or any later skill installation/update. The managed marker is only checked during pruning/cleanup, after the destructive write has already happened.

Please make command generation ownership-safe:

- create the file when absent;
- update it only when it already contains the exact AI Factory marker;
- otherwise preserve the user file and report a conflict, or use a
  collision-free naming strategy.

Please add a regression test that pre-creates a user-owned `aif.md` and verifies that init/update leaves it unchanged.

### P2 — A removal-only update leaves stale managed command stubs

The pruning block at `src/core/transformers/zcode.ts:58-67` lives in `postInstall()`, but `updateSkills()` reaches `postInstall()` only through `installSkills()`.

When a skill is removed from the package and no other skill is new or needs reinstalling, `updateSkills()` removes the skill directory but never invokes `postInstall()`. The old `.zcode/commands/<removed-skill>.md` therefore remains and points to a missing skill, contrary to the PR's stated pruning behavior.

Run ZCode command reconciliation once per agent update even when the install set is empty, or remove the matching managed stub in the package-removal path. Please add a removal-only regression test.

### Verification notes

- Reviewed all 11 changed files.
- The ZCode workspace paths and `mcp.servers` nesting match the current ZCode documentation.
- The current GitHub Actions run has conclusion `action_required` and contains  zero jobs, so CI has not executed for this head yet.
- This PR adds no automated ZCode tests. The stated `npm test` result exercises the skill-validation suite, not these transformer/MCP lifecycle branches.

PR #150, opened 2026-08-20
feat(skills): add session warmup workflow

lee-to:
## Summary

- add `/aif-warmup` to load the core AI Factory artifacts and `AGENTS.md` into a compact, fork-ready session handoff
- add the user-owned `warmup.paths` configuration for extra project-root files and directories, with safe text-only traversal and clear skip reporting
- register the skill as a workflow across supported transformers and document and test the new configuration contract

## Testing

- `npm run build`
- `npm test` (147 passed)
- `bash scripts/test-aif-config.sh`
- transformer output checks for Antigravity and KiloCode

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,7 @@
 **AI Factory** (v2) is an npm package + skill system that automates AI agent context setup for projects. It provides:
 
 1. **CLI tool** (`ai-factory init/update/upgrade`) — installs skills and configures MCP
-2. **Built-in skills** (29 skills, all `aif-*` prefixed) — workflow commands for spec-driven development
+2. **Built-in skills** (30 skills, all `aif-*` prefixed) — workflow commands for spec-driven development
 3. **Spec-driven workflow** — structured approach: plan → implement → commit
 4. **Multi-agent support** — 16 agents (Claude Code, Cursor, Windsurf, Roo Code, Kilo Code, Antigravity, OpenCode, Warp,
    Zencoder, Codex CLI, Codex app, GitHub Copilot, Gemini CLI, Junie, Qwen Code, Universal)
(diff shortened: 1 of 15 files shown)


ichinya (commented):
Clean, focused addition. The new `/aif-warmup` skill is well-scoped: read-only, config-aware, fork-ready, and documented consistently across `AGENTS.md`, `docs/*`, `config-template.yaml`, and the transformer / wizard code. The tests in `scripts/test-aif-config.sh` and `scripts/test-skills.sh` cover the new contract.

**Highlights**
- Good use of `disable-model-invocation: true` for a context-only loader, consistent with `aif-grounded` / `aif-explore`.
- The user-owned `warmup.paths` design is clean: `/aif` only preserves it, never overwrites it.
- Transformer and hint registration look correct (`WORKFLOW_SKILLS`, Antigrativity bullet list, `skill-hints.ts`).

**Nits / questions (non-blocking)**

1. `skills/aif/references/update-config.mjs` does not include `warmup` in `SECTION_KEYS` / `ALLOWED_PATHS`. I understand why it cannot be a "managed" key — the value is a YAML list (`[]`) and the helper rejects `[` scalars. In `create` mode the block comes from the template, and in `merge` mode the unknown section is preserved, so the tests pass. But because `SECTION_ORDER` does not include `warmup`, a missing managed section inserted during `merge` (e.g. `workflow`) could be placed at the start of the `warmup` comment block and push `warmup` down. Is this ordering acceptable, or should `warmup` become an insertion anchor in `SECTION_ORDER` without managed keys?

2. In `skills/aif-warmup/SKILL.md` Step 2 the rules read order is "top-level rules file, `rules.base`, then every named `rules.<area>`", while the priority is `rules.<area>` > `rules.base` > `paths.rules_file`. Consider aligning the wording so the read/priority relationship is unambiguous.

3. `/aif-warmup` is grouped under "Core workflow and quality" in `AGENTS.md` and `docs/configuration.md`. It feels more like an additional utility skill. If that was a deliberate grouping, feel free to ignore.

No blockers — overall LGTM.

PR #154, opened 2026-09-02
feat(aif-qa-check): persist browser replay scripts

lee-to:
## Summary

- persist one dependency-free Playwright-compatible browser replay script per QA case
- reuse replay only when case digest, approved-target fingerprint, and last proven script-content digest all match
- generate new scripts before their first run and use that single execution as proof instead of automatically running stateful scenarios twice
- preserve original scripts and errors before repair, and treat missing selectors as product failures unless semantic equivalence is proven
- bind navigation to the approved runtime target through AIF_BASE_URL and require renewed authorization for repeated side effects
- replay safe prior passes and failures after code/build changes, then make a recorded risk-based decision about bounded exploratory coverage

## Testing

- bash scripts/test-aif-qa-check.sh (20/20 passed)
- bash skills/aif-skill-generator/scripts/validate.sh skills/aif-qa-check (passed, no warnings)
- npm test with an isolated temporary npm cache (147/147 passed)


ichinya:
The idea of using persistent browser replay artifacts is useful, but the current contract is not yet safe to use as a regression baseline. I found four blocking issues.

### 1. [P1] Do not automatically replay a new or updated replay a second time

In the `Browser Replay Contract`, clause 4, for a missing or outdated script, it is prescribed to first execute the case when creating the script, and then immediately execute the saved script again.

For stateful scenarios, this means two side effects in a row:

* creating a user or an order;
* sending an email or a notification;
* changing permissions;
* changing settings;
* deleting a record;
* submitting a form.

The second run may repeat an external action or cause a false failure because the first run has already changed the state. The existing per-case authorization also does not inform the user that the action will be performed twice.

It is better to generate a script before the first execution and consider the first run as a test. If the steps had to be clarified interactively, re‑running is permissible only after explicit restoration of preconditions/cleanup. For side‑effect cases, separate confirmation of re‑execution is also required.

### 2. [P1] Selector failure cannot be automatically considered an automation defect

Currently, matching script with `selector, syntax, or runner failure` is proposed to be repaired, and it is explicitly prohibited to classify such an error as a product regression.

This is reasonable for `syntax` and `runner`. However, a missing selector is often an observable regression:

* the component didn’t render;
* the button disappeared;
* the page entered an incorrect state;
* permissions changed;
* the user ended up on an unexpected route;
* the required element was replaced by another element with different behavior.

Automatic “repair” may select a similar element, bypass the broken step, or modify the check in such a way that the defect is hidden.

For selector failure, you first need to prove that the same semantic element and the expected behavior have been preserved, and only the locator mechanics have changed. If this is not proven, the case should receive the `Failed` status.

The original script and the initial error should be saved as evidence, rather than silently rewritten.

### 3. [P1] Replay must be tied to the current target environment

The script must include navigation, but its relevance is determined only by `case_digest`. The digest itself does not contain a URL, origin, or environment, and replay is deliberately reused between:

* revision changes;
* worktree changes;
* manual build changes.

Because of this, a script created for an old preview or staging URL will remain `matching` when switching to another worktree, a new preview, or a local build, and will be run first.

As a result, you may:

* check the wrong build;
* get a false result;
* perform actions in the old environment;
* Cross the safety boundary, despite the target environment shown to the user during preflight.

One of the following options is required:

1. A separate target fingerprint/origin in the script’s metadata, with mandatory verification before execution.
2. Relative routes with a runtime‑injected base URL.
3. An explicit placeholder like `AIF_BASE_URL`, resolved before execution.

If the target does not match, the script cannot be executed until rebinding or manual verification.

### 4. [P2] We need to link QA evidence to the contents of the replay script

The entire `browser_replay_dir` is excluded from `worktree_digest`, and the script’s validity is checked only by the first line:

```js
// aif-case-digest: <case_digest>
```

If in `TC-001.js` to change the URL, actions, or remove an assertion while keeping the same `case_digest`, the existing QA results will not become stale, and the modified script will continue to be considered matching.

The same will happen after automatic repair of the replay file.

It is worth calculating a separate content digest for the replay script, recording it in `qa-check.md`, and verifying it before reusing:

```text
Browser replay:
  Path: browser-replay/TC-001.js
  Case digest: ...
  Script digest: ...
  Target fingerprint: ...
  Execution capability: ...

Changing or repairing a script must update the `script_digest` and require a repeatable proof run.

`case_digest` is responsible for ensuring the script matches the test case, while `script_digest` ensures the invariability of the actually executed check.

After addressing these points, the approach itself looks reasonable: a separate file per case, reusing previously passed/failed scenarios, and bounded exploration complement the existing revision/worktree binding well.

ichinya (changes requested):
All four observations follow directly from the added replay contract and the exclusion of browser_replay_dir from the digest of the working tree.

ichinya:
In ichinya/aifhub-extension, there is already a close regression-first contract:

* the same regressionCheck is performed before and after the fix;
* for implementation, the RED → GREEN → REFACTOR cycle is used;
* testCheck, redResult, greenResult, refactorResult are saved, and for fix — rootCauseEvidence, hypothesis, experiment, preFixResult and postFixResult;
* the results remain supporting evidence, and authoritative verification belongs to /aif-verify.

This is not an objection to the PR — on the contrary, it is logical to transfer such general mechanics from the extension to the core AI Factory solution. When implementing this, you can use the existing AIFHub Extension contract as an additional reference and, where applicable, standardize the terminology and evidence semantics so that the core and the extension do not develop two similar but gradually diverging approaches.

After the corresponding core contract is available, I will be able to simplify the aifhub-extension on my part: remove duplicate general instructions and keep only the OpenSpec‑specific integration, runtime evidence paths, and adaptation to the extension’s lifecycle.

Relevant implementations:

* ichinya/aifhub-extension#132 — regression‑first fix evidence;
* ichinya/aifhub-extension#179 — RED → GREEN → REFACTOR, systematic debugging and supporting traces.

lee-to:
Addressed all four blocking findings in commit a9375d1:

1. New or updated replay scripts are created before execution and run once as proof. There is no automatic second execution; stateful repeats require restored/verified preconditions and renewed per-case authorization.
2. Selector failures are now possible product regressions. Locator repair is allowed only with concrete evidence that the same semantic element and expected behavior remain; the original script and initial error/state are preserved first.
3. Navigation uses the runtime-bound AIF_BASE_URL placeholder. Every replay verifies an approved-target fingerprint before execution, and target rebinding requires a new proof run.
4. QA evidence now records the exact proven script-content digest. Script changes stale the result; replaced scripts are archived under browser-replay/history/.

I also reviewed the referenced aifhub-extension regression-first work and kept this core contract compatible with the existing AI Factory evidence semantics without copying the extension-specific lifecycle.

Validation: qa-check contract 20/20, skill validator passed with no warnings, full suite 147/147.

lee-to:
All requested review changes have now been implemented and pushed in commit a9375d1. The full local suite passes (147/147), CI is green, and the PR is ready for re-review. Thank you for the thorough review.

ichinya:
## Re-review after a9375d1

Rechecked the complete diff and the follow-up commit. All four blocking findings from the previous review are now addressed coherently.

### Previous findings

1. **No automatic duplicate execution**

   Missing or changed scripts are created before execution, and their first execution becomes the proof run. A new or updated stateful scenario is no longer automatically executed twice.

   Repeating a case that can create, submit, send, delete, change permissions/settings, or otherwise produce an external side effect now requires:

   - restored or explicitly verified preconditions;
   - renewed per-case authorization;
   - an explicit indication that the action is being repeated.

2. **Selector failures remain possible product regressions**

   Missing selectors are no longer classified automatically as automation defects.

   Locator repair is allowed only when concrete browser evidence proves that:

   - the same semantic element still exists;
   - the expected behavior is unchanged;
   - only the locator mechanics changed.

   Otherwise, the case is correctly recorded as `Failed`. The original script and initial error/state are preserved before any repair.

3. **Replay is bound to the approved execution target**

   Persisted scripts now use the runtime-bound `"AIF_BASE_URL"` placeholder rather than an absolute environment URL.

   The separate `target_fingerprint`, derived from the normalized target and environment class, prevents a script created for one preview, staging instance, or local environment from being silently reused against another target.

   Target rebinding changes the replay binding and requires a new proof run.

4. **QA evidence is bound to the exact replay content**

   `script_digest` is now separate from `case_digest`.

   The resulting identity is:

   ```text
   case_digest
   + target_fingerprint
   + script_digest

lee-to:
Thanks @ichinya for the thorough re-review and for confirming that all four blocking findings are now addressed coherently.

I also verified that the PR branch is clean and CI is green. The re-review comment appears to end in the middle of the replay identity code block after case_digest + target_fingerprint + script_digest. If any additional feedback was omitted, please add it and I will address it.

Otherwise, the PR still shows the original CHANGES_REQUESTED review state, so could you please update it to approved when convenient?

ichinya (approved):

PR #162, opened 2026-09-16
fix(installer): respect aif-loop selection for bundled subagents

### LINKED ISSUE #161: всегда устанавливается /aif-loop даже если его не выбрать при установке

lee-to:
## Summary

Selecting skills during `ai-factory init` previously had no effect on bundled subagents: deselecting `aif-loop` still installed every `loop-*` helper. Rerunning `init` also left previously installed, deselected base skills on disk.

- Require `aif-loop` before installing bundled `loop-*` subagents for Claude and Codex.
- Remove previously tracked loop subagents and their metadata when the skill is deselected, and keep them excluded during `update`, including `--force`.
- Remove deselected base skills on `init` reruns only when the saved raw ownership receipt verifies unchanged managed contents. Preserve modified or unproven files with a warning, including user files inside tracked skill directories and flat workflows, and stop tracking the deselected skill.
- Preserve custom skills, untracked files, and extension replacements.
- Document the selection behavior and add smoke coverage for fresh installations, reselection, cleanup, legacy metadata, and preservation of user/extension-owned files.

The current Codex bundle has no loop roles; isolated TOML fixtures verify that the same selection rule applies if they are added.

Closes #161.

## Validation

- `npm run build`
- `npm run lint` — TypeScript unused checks and Knip both pass; the loop smoke test has an explicit npm entry point.
- `npm --cache /private/tmp/aif-npm-cache test` — all 154 checks passed, including init/update smoke tests and Claude/Codex loop selection coverage.
- Confirmed the new regression check fails before the subagent selection fix.
- `node scripts/test-managed-skill-receipts.mjs` — covers clean removal, extra user files before/after installation, modified files, missing raw receipts, and missing managed state across Claude, Codex, Antigravity, and Kilo Code. The preservation scenarios fail before the ownership fix and pass afterward.
- `npm run test:cleanup-helper`


ichinya:
## Review: Request changes

Reviewed commit: 9344031110c76cadca02bc5c1dcc9101a90c809b.

Keeping bundled inventory discovery separate from skill-selection filtering is a good approach. However, the new base-skill cleanup introduces a data-loss path that should be addressed before merging.

### [P1] Verify ownership before deleting deselected skill directories

**Location:** [src/cli/commands/init.ts:239–249](https://github.com/lee-to/ai-factory/blob/9344031110c76cadca02bc5c1dcc9101a90c809b/src/cli/commands/init.ts#L239-L249)

The new `removeSkillsByName(projectDir, existingAgent, deselectedSkills)` call treats membership in `installedSkills` as sufficient permission to delete the entire installed directory. That helper performs recursive removal without checking `managedSkills`, the saved `rawInstalledHash`, or the current contents.

A tracked base skill can still contain user-owned files. For example, the initial installation can overlay an existing skill directory while leaving a user's additional files intact. `buildManagedSkillsState()` explicitly distinguishes this case: it can record the observed installation while withholding the raw ownership receipt when the contents cannot be proven to come from the managed source. This new cleanup path ignores that distinction.

A regression scenario using the CLI built from this PR is:

```sh
# Point this at the built PR checkout, not a globally installed release.
AIF_CLI=/absolute/path/to/ai-factory/dist/cli/index.js
PROJECT_DIR="$(mktemp -d)"
cd "$PROJECT_DIR"

mkdir -p .claude/skills/aif-loop
printf 'User-owned notes\n' > .claude/skills/aif-loop/personal-notes.md

node "$AIF_CLI" init --agents claude --skills aif,aif-loop
node "$AIF_CLI" init --agents claude --skills aif

# Must remain present: AI Factory never installed this file.
test -f .claude/skills/aif-loop/personal-notes.md
```

The new deletion path removes the whole `aif-loop` directory, including `personal-notes.md`. The same issue applies to local edits and installations without a verifiable raw ownership receipt. Testing preservation of a sibling `custom-skill` directory does not cover user-owned content inside a tracked base-skill directory.

Please reuse or extract the ownership/content verification already implemented by `removeOwnedSkills()`, restricted to the deselected skills. Delete only provably managed, unchanged artifacts; preserve and warn about modified or unproven content. Also distinguish “preserved for safety” from an actual removal failure: the new `failedSkills` check would otherwise turn intentional preservation into an error.

Regression coverage should include clean managed removal, an additional user-owned file inside a tracked skill, a modified skill file, and a missing raw receipt. The preservation rule should also apply to flat workflow files.

### CI validation needs to be completed

The [CI run for this head commit](https://github.com/lee-to/ai-factory/actions/runs/35100977339/job/104810033248) passed **Build** but failed **Lint**. **Validate skills**, **Init smoke tests**, **Update smoke tests**, and **Cleanup helper tests** were consequently skipped. The [base commit's check](https://github.com/lee-to/ai-factory/actions/runs/34747027074/job/103696720112) was successful.

The PR description reports `npm run lint:unused`, but the repository's full `npm run lint` also runs `lint:dead` (`knip --reporter compact`). Please run the complete lint command and obtain a green CI run so that the smoke tests actually execute. The exact lint diagnostic was not available through the connector, so I am not attributing the failure to a specific rule or file.

**Verification scope:** source/diff review, GitHub check results, and an isolated filesystem probe of the transcribed cleanup logic. The probe confirmed deletion of an additional user-owned file despite a missing raw ownership receipt. I did not run the full repository test suite or the end-to-end CLI regression scenario above locally.

lee-to:
Addressed both review points in c15504257a78e5e7ea7b7f472162ce538a0077cc.

### Ownership-safe cleanup

`init` now calls the existing `removeOwnedSkills()` path, restricted to deselected skills. Deletion requires the saved ownership state and raw installed hash to match the current contents. Modified or unproven artifacts are preserved with a warning, including extra user files inside tracked directories and flat workflow files.

Intentional preservation no longer triggers the `failedSkills` error. The deselected skill is removed from CLI tracking, while actual filesystem errors still propagate from the removal helper.

Added CLI regression coverage for clean managed removal, user files present before installation or added afterward, modified skill files, missing raw receipts, and missing managed state. The cases cover Claude/Codex skill directories and Antigravity/Kilo Code flat workflows. The preservation regressions fail on the previous implementation and pass with this fix.

### Lint and validation

The lint failure was Knip reporting `scripts/test-loop-subagents.mjs` as unused because its invocation was inside a shell script. Added an explicit `test:loop-subagents` npm entry point, so Knip recognizes the test without suppressing the check.

Local validation passed:

- `npm run build`
- `npm run lint` (including `lint:dead`)
- `npm test` with a writable temporary npm cache: 154/154 checks, including init/update smoke tests
- `npm run test:cleanup-helper`: 30/30 checks

[GitHub CI for c155042](https://github.com/lee-to/ai-factory/actions/runs/35361174151) is green. Build, full Lint, Validate skills, Init smoke tests, Update smoke tests, and Cleanup helper tests all executed successfully.

